PR #102, opened 2013-06-12
[COOK-3155] Be able to install chef-client from omnibus packages

guilhem:
https://tickets.opscode.com/browse/COOK-3155

### DIFF
--- a/attributes/install.rb
+++ b/attributes/install.rb
@@ -1,0 +1,2 @@
+default['chef-client']['version']        = :latest
+default['chef-client']['install_method'] = "omnibus"
--- a/attributes/install_omnibus.rb
+++ b/attributes/install_omnibus.rb
@@ -1,0 +1,7 @@
+include_attribute "chef-client::install"
+
+default['chef-client']['omnibus']['version'] = node['chef-client']['version']
+default['chef-client']['omnibus']['prereleases'] = false
+default['chef-client']['omnibus']['nightlies'] = false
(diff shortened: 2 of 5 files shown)


pburkholder:
More on this, to quote someone knowledgeable on the history here, regarding why chef-client doesn't support updates:

First, the `chef-client` cookbook concerns itself with the configuration of the `chef-client` program. When building cookbooks it is generally advised to keep their focus narrowed to a specific component. This means that for `chef-client` we're mainly concerned with the configuration (/etc/chef/client.rb), and running it regularly, be that as a service or a cron job.

Second, the cookbook predated the creation of omnibus, and at one point attempted to handle upgrading the chef gem. [snip]. When omnibus came along, the upgrade process became more complex. We now have an API service that concerns itself with determining the right package file to retrieve for installation, but we (CHEF) never really wrote a client to handle that because we had the `install.sh` script do it.

Heavy Water created the omnibus-updater cookbook to cover this gap. It works very well. At this point in time there is probably enough adoption of that cookbook that converting people to use something else is like turning a barge in a river. 

Migrating the functionality of the omnibus-updater cookbook to the chef-client cookbook would make the latter's codebase more complicated (it's already reasonably complex) and potentially difficult for users and contributors to follow. It would require more effort in testing and we simply don't have the bandwidth for that. Heavy Water is doing a good job of maintaining that cookbook and I'm inclined to keep it that way, rather than try to absorb that into our cookbook

PR #127, opened 2013-11-07
[COOK-3972] When using chef-client::cron, allow the option to append output instead of overwriting.

mattchukabam:
The default of overwriting is preserved, but this is useful for CI
environments where a bad push might not get caught until well after the
fact (most importantly, after the next automated chef run, where
previous logs would be destroyed).

The actual impl is admittedly simple, and makes it slightly harder to
read; suggestions are welcome.

Jira ticket: https://tickets.opscode.com/browse/COOK-3972


juliandunn:
Closing in favour of #261 due to non-response.

PR #132, opened 2013-12-05
optionally append to the chef log in cron mode

otakup0pe:
What it says on the tin. Default behavior remains the same - log will be overwritten each time.

This was my first time doing test kitchen things - let me know if it's appropriate.


juliandunn:
Closing in favor of #261 due to non-response

PR #141, opened 2013-12-31
handle nil client key

spheromak:
In some work I was doing with packer and client-zero I ran into a case with a nil `Chef::Config[:client_key]`. This adds a default. I know this isn't proper for windows, maybe someone has input on the secret sauce for where the default should be on that platform. 

Tried to make an issue for this in tickets.opscode.com, but my login is wonked, and I can't reset it for whatever reason.


jtimberman:
@spheromak Any update on this? Is it still an issue? Thanks!

PR #187, opened 2014-09-02
Handle verify_api_cert parameter by copying trusted certificate file to ...

snkutlu:
Handle verify_api_cert parameter by copying trusted certificate file to client before client.rb is updated

Trusted certificates and verify_api_cert is handled with knife bootstrap command during initial bootstrap but for existing clients, this change will copy trusted certificate of Chef server from URI in node['chef_client']['trusted_cert'] to /etc/chef/trusted_certs directory if 
- verify_api_cert attribute is set to true and 
- node['chef_client']['trusted_cert'] attribute is not empty

If verify_api_cert is true and 'trusted_cert' attribute is not set, then in order not to break SSL communication, code will set verify_api_cert to false for that chef-client session.

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -42,6 +42,9 @@ default['chef_client']['splay']       = '300'
 default['chef_client']['conf_dir']    = '/etc/chef'
 default['chef_client']['bin']         = '/usr/bin/chef-client'
 
+default['chef_client']['trusted_certs_dir'] = "#{node['chef_client']['conf_dir']}/trusted_certs"
+default['chef_client']['trusted_cert'] = ""
+
 # Set a sane default log directory location, overriden by specific
 # platforms below.
 default['chef_client']['log_dir']     = '/var/log/chef'
--- a/metadata.rb
+++ b/metadata.rb
@@ -4,7 +4,7 @@ maintainer_email  'cookbooks@opscode.com'
 license           'Apache 2.0'
 description       'Manages client.rb configuration and chef-client service'
 long_description  IO.read(File.join(File.dirname(__FILE__), 'README.md'))
-version           '3.7.0'
+version           '3.7.1'
 recipe 'chef-client', 'Includes the service recipe by default.'
 recipe 'chef-client::arch_service', 'Configures chef-client as a service on Arch Linux'
 recipe 'chef-client::bluepill_service', 'Configures chef-client as a service under Bluepill'
--- a/recipes/config.rb
+++ b/recipes/config.rb
@@ -74,7 +74,7 @@ d_group = dir_group
 # If verify_api_cert is defined to be true, 
 # then we need Chef server certificate in trusted certificate directory
 #
-if node['chef_client']['config']['verify_api_cert'] == "true"
+if node['chef_client']['config']['verify_api_cert'] == true
   # If trusted certificate filename attribute is not defined
   # set verify_api_cert to false to prevent SSL certificate errors
   if node['chef_client']['trusted_cert'] != ""
--- a/recipes/config.rb
+++ b/recipes/config.rb
@@ -70,6 +70,35 @@ end
 d_owner = dir_owner
 d_group = dir_group
 
+#
+# If verify_api_cert is defined to be true, 
+# then we need Chef server certificate in trusted certificate directory
+#
+if node['chef_client']['config']['verify_api_cert'] == "true"
+  # If trusted certificate filename attribute is not defined
+  # set verify_api_cert to false to prevent SSL certificate errors
+  if node['chef_client']['trusted_cert'] != ""
+    directory node['chef_client']['trusted_certs_dir'] do
+      recursive true
+      owner d_owner
+      group d_group
+      mode 00755
+    end
+    cert_filename= File.basename(node['chef_client']['trusted_cert'],".*")
+    remote_file "trusted_certificate" do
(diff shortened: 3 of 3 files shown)


jtimberman:
I'm generally :-1: on this functionality, because I think this sort of thing should happen out of band of the Chef Client run, like with `bootstrap` methods (knife, chef provisioning), because we may have a situation where chef-client cannot connect to the Chef Server to even get this cookbook in the first place.

That said, if there's consensus from others to move forward with this, I have some additional thoughts about implementation.

We should do the cert retrieval in a separate, new recipe. I don't think it should be included by default or using an attribute, because we already have enough attribute cognitive load in this cookbook. I don't think we should be setting an override attribute in the recipe (the `else` fall back that sets `verify_api_cert` to false). Nor do I think we should set that to false at this level, and instead defer that to individual site policy. If your local policy says to disable it, then you set that attribute in your role/wrapper cookbook or wherever you set attributes.

There is already a `Chef::Config[:trusted_certs_dir]` configuration option, and we should rely on that rather than introduce `['chef_client']['trusted_certs_dir']`. If this attribute *is* required here, we should set it with the `['chef_client']['conf']` hash.

We still need a CLA on file for @snkutlu. Sometimes we can take the work done in a PR and add it ourselves rather than wait, but we have two people that aren't totally in favor of this change (myself and @juliandunn, dunno if others?).

Finally, please don't increment the version in the metadata. We do that upon release. This is likely contributing to this pull request not being able to merge cleanly.

danielsdeleo:
My only opinion here is that we should help users upgrade to enabling SSL certificate verification in any way we can. I'll leave it up to everyone else to balance the trade-offs of different ways of implementing that, since I don't have nearly the amount of expertise in this code base that y'all have.

juliandunn:
After much discussion I don't think we're going to do this.

acqant:
I'm really close to turning SSL validation off on chef 12.  I'm another user who is behind a corp SSL firewall and embedded cacert.pem without my corp certs gets in the way of a lot.

Writing something to manage the appendage of certs to chef's embedded cert bundle is a hassle.  Is this the only option?  Is their nothing in test-kitchen knife bootstrap that can take a payload of additional ssl certs to install?


@juliandunn @danielsdeleo @jtimberman

danielsdeleo:
@acqant I'm confused by this statement:

> Is their nothing in test-kitchen knife bootstrap that can take a payload of additional ssl certs to install?

Test Kitchen and knife bootstrap are different things. Are you saying that neither of them work? I definitely think we may have forgotten to add trusted certs support to kitchen because internally we tend to use the chef-zero provisioner(s). If you're using the client/server provisioner and it doesn't support trusted certs, it's definitely worth filing an issue at https://github.com/test-kitchen/test-kitchen/issues

As for `knife bootstrap`, it already copies your trusted certs directory to the bootstrap host, from there it should be used for all SSL certificate verification.

acqant:
My head is swimming a bit from research of "trusted_certs" and
"ssl_ca_path" so I'll tell you what I'm trying to do and what is failing.

Brand new environment, I'm only working with test-kitchen and chef_zero at
the moment, no chef server yet.
Unless I append my corporate SSL intermediate certs to
/opt/chef/embedded/ssl/certs/cacert.pem anything that requires SSL and uses
embedded chef fails.

I forget what, exactly, failed.  Probably an embedded chef gem or web
resource.

danielsdeleo:
@acqant "trusted_certs" is the feature we're (tangentially) discussing in this issue. It's a feature where you can put any certificates in a directory called `trusted_certs` ( for knife, `~/.chef/trusted_certs`, for chef-client, `/etc/chef/trusted_certs`) and Chef's HTTP stack will add these to the OpenSSL certificate store before starting an HTTPS request.

acqant:
awesome... that's what I thought and hoped.  I went looking for
"test-kitchen and trusted_certs" and didn't find much.

I did find this when I smacked into the same issue with sensu and this
might also work in the short term for test-kitchen.  I'll try dropping my
certs into
/opt/chef/embedded/lib/ruby/site_ruby/2.1.0/rubygems/ssl_certs/

https://github.com/sensu/sensu-community-plugins/issues/324

danielsdeleo:
I think only rubygems will use certs in that directory. Rubygems started bundling its own CA certs (basically the same thing we did for Chef) at some point.

acqant:
​So the only real issue is test-kitchen doesn't expose trusted_certs ???

danielsdeleo:
For your use case, this might be true. If you manually add the required certs to /etc/chef/trusted_certs, then the expected behavior is that HTTPS requests made by chef (`http_request`, `remote_file`, requests to a Chef Server) should work. Is that the case?

acqant:
I'd have to edit the /tmp/kitchen/config.rb and run the chef-client to
see.  I'm still trying to remember what besides gem installs failed.

acqant:
Can confirm that adding trusted_certs_dir to config.rb in /tmp/kitchen and
running

sh -c '
sudo -E /opt/chef/bin/chef-client --local-mode --config
/tmp/kitchen/client.rb --log_level info --force-formatter --no-color
--json-attributes /tmp/kitchen/dna.json --chef-zero-port 8889
'
gets resources like 'remote_file' to work in test-kitchen with SSL
intermediate certs.

danielsdeleo:
Cool, can you file an issue for test-kitchen for this?

PR #189, opened 2014-09-05
Allow blank log_dir

gh2k:
Doesn't override log location with -L parameter when log_dir is not set, meaning we can override log_location in client.d

Currently if I set log_location (for example, to log to syslog) it silently has no effect due to this command-line parameter.


gh2k:
I cleaned up the conflicts again, but it looks like #219 may well have made this redundant? @kamaradclimber would you agree?

kamaradclimber:
yes I think so

gh2k:
Thanks. Closing this now.

PR #194, opened 2014-09-09
Fix FC023 violation:  Prefer conditional attributes

nathenharvey:


jtimberman:
I made a decision :cake:.

PR #203, opened 2014-09-25
correctly set "chef_server_user" for Chef Server 12

JeanMertz:
Granted, this isn't the best solution, but Chef::VERSION was returning `11.x.x` because I am still using the 11.x version of chef-client.

This probably won't be merged, but at least others that want to setup Chef server 12 rc can use it for now.

### DIFF
--- a/libraries/helpers.rb
+++ b/libraries/helpers.rb
@@ -24,7 +24,11 @@ module Opscode
       include Chef::DSL::PlatformIntrospection if Chef::VERSION >= '11.0.0'
 
       def chef_server_user
-        Chef::VERSION >= '11.0.0' ? 'chef_server' : 'chef'
+        if Chef::VERSION >= '11.0.0'
+          system('getent passwd opscode') ? 'opscode' : 'chef_server'
+        else
+          'chef'
+        end
       end
 
       def chef_server?


jtimberman:
The chef-server cookbook was [updated](https://github.com/chef-cookbooks/chef-server/commit/0f2d123ad9ebb40ac18fdabdeee2d66735604bbe) for Chef Server 12 a few weeks ago.

libraries/helpers.rb
```diff
@@ -24,7 +24,11 @@ module Helpers
       include Chef::DSL::PlatformIntrospection if Chef::VERSION >= '11.0.0'
 
       def chef_server_user
-        Chef::VERSION >= '11.0.0' ? 'chef_server' : 'chef'
+        if Chef::VERSION >= '11.0.0'
+          system('getent passwd opscode') ? 'opscode' : 'chef_server'
+        else
+          'chef'
+        end
```
jtimberman: The disconnect here is that this checks that the chef client version is 11.

However, I question whether this is actually necessary with Chef Server 12? It keeps all its directories separate and the `chef-server-ctl` command manages those with its internal cookbooks. This is also true of Chef Server 11..

jtimberman:
To follow up to the earlier comments, I think we probably want #253, and remove all this code.

jtimberman:
Actually #253 is implemented in #286, which also updated the specs. I think we should just have that code removed so we don't have to maintain what was present for older now-unsupported Chef Server versions.

PR #207, opened 2014-10-03
Added support to use the config recipe with upstart and logrotate

brodock:
I've changed chef-client::config to create a specific logrotate for when using upstart init system. As a fallback for when init system can't be detected, it will do a copytruncate.

### DIFF
--- a/recipes/config.rb
+++ b/recipes/config.rb
@@ -30,6 +30,18 @@ case node['chef_client']['log_file']
 when String
   log_path = File.join(node['chef_client']['log_dir'], node['chef_client']['log_file'])
   node.default['chef_client']['config']['log_location'] = "'#{log_path}'"
+  logrotate_options = ['compress']
+
+  case node['chef_client']['init_style']
+  when 'init'
+    logrotate_postrotate = '/etc/init.d/chef-client reload >/dev/null || :'
+  when 'upstart'
+    logrotate_postrotate = 'restart chef-client >/dev/null || :'
+  else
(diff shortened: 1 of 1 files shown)


brodock:
does it need anything more to get merged?

jtimberman:
I think it would be cleaner to delegate this to a library helper method.

jhh:
While you're at it, can you add support for `systemd`? My centos 7 system is getting cron errors from having `init` style configured. Something like:

```ruby
  case node['chef_client']['init_style']
  when 'init'
    logrotate_postrotate = '/etc/init.d/chef-client reload >/dev/null || :'
  when 'upstart'
    logrotate_postrotate = 'restart chef-client >/dev/null || :'
  when 'systemd'
    logrotate_postrotate = 'systemctl reload chef-client.service > /dev/null || :'
```

juliandunn:
There were other changes to logrotate merged in #241 and #243 - can you see if these are sufficient?

jtimberman:
@jhh We have systemd support:

```
7 weeks ago ea0a955 Allan Espinosa  Switch Arch Linux to use systemd.
6 months ago a44c248 Julian C. Dunn  Allow EL7 to use the systemd recipe
9 months ago b963527 Julian C. Dunn  [COOK-3465] Switch Fedora to using systemd
```

jtimberman:
Closing, as #241 and #243 will cover the functionality required, and give users configurable knobs in the form of attributes.

PR #232, opened 2014-12-17
ignore empty strings in the client.rb file

dpetzel:
This PR is has couple loosely related objecives.

In an environments where *almost* all your nodes use a proxy server, you might set your proxy configuration in your base role or environment file. You then realize having your proxy servers configured to use themselves as proxy servers may not be ideal.... The logical solution is to set an attribute override to wipe out the default that you setting. Sadly if you try that (like I did), you run into the fact that setting a nil value in an override doesn't actually work: https://github.com/opscode/chef/issues/1507

No problem, I'll set the values to an empty string... Sadly that results in the values getting written to the client.rb as an empty string:(

Inspired by this https://github.com/miketheman/nginx/commit/b21cb4ed14578185f4479ce195f986f8f783aff0 this PR will update the client.rb template to not write out configuration elements that are not empty strings. I couldn't think of any situation where it would be valid do to so, but I admit, its not impossible there could be one. We then take that some approach for the proxy ENV vars, and don't write them out if the strings are empty.

The last update to the template is to use the `@chef_config` when evaluating the proxy settings rather than the node attribute. This is for consistency more than anything else. The `@chef_config` already has the data that was passed in, so we can access the config in a uniform fashion.

Lastly, a new test context for ensuring the empty strings don't get written out.

### DIFF
--- a/spec/unit/config_spec.rb
+++ b/spec/unit/config_spec.rb
@@ -116,6 +116,30 @@ describe 'chef-client::config' do
       expect(template).to_not notify('ruby_block[reload_client_config]')
     end
 
+    context 'empty config key values' do
+      let(:chef_run) do
+        ChefSpec::Runner.new do |node|
+          node.set['chef_client']['config']['http_proxy'] = ''
+          node.set['chef_client']['config']['https_proxy'] = ''
+          node.set['chef_client']['config']['no_proxy'] = ''
+        end.converge(described_recipe)
+      end
+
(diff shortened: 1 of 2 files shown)


juliandunn:
Seems ok to me, but needs a rebase. Sorry for the delay in reading this.

juliandunn:
Assuming abandoned since no activity in a while. Please reopen if you intend to still work on this.

PR #238, opened 2015-01-07
http_proxy, https_proxy & no_proxy options to set ENV variables

willejs:
There is no need to set ENV variables in the client.rb, instead they are standard options.
This might be a new change to the config and this cookbook doesn't install a specific version of chef-client, so I am unsure if instead i should put some logic in here?

Obvious fix.

### DIFF
--- a/templates/default/client.rb.erb
+++ b/templates/default/client.rb.erb
@@ -30,7 +30,7 @@ node_name <%= @chef_config['node_name'].inspect %>
 http_proxy "<%= node["chef_client"]["config"]["http_proxy"] %>"
 <% end -%>
 <% unless node["chef_client"]["config"]["https_proxy"].nil? -%>
-https_proxy = "<%= node["chef_client"]["config"]["https_proxy"] %>"
+https_proxy "<%= node["chef_client"]["config"]["https_proxy"] %>"
 <% end -%>
 <% unless node["chef_client"]["config"]["no_proxy"].nil? -%>
 no_proxy "<%= node["chef_client"]["config"]["no_proxy"] %>"
--- a/templates/default/client.rb.erb
+++ b/templates/default/client.rb.erb
@@ -24,19 +24,16 @@ node_name <%= @chef_config['node_name'].inspect %>
 <% else -%>
 # Using default node name (fqdn)
 <% end -%>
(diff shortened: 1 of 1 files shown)


juliandunn:
I believe the reason the proxy environment variables are set is indicated in this comment: https://github.com/opscode-cookbooks/chef-client/pull/238/files#diff-3a090d01433d142e84af2c33eeab49b5L29

juliandunn:
@willejs actually I tested this with a server that can only reach the Internet via a proxy and it works fine, so I can take your patch. Can you sign the CLA please?

willejs:
@juliandunn thanks for looking at this. I just want to double check something before you merge.

willejs:
Right, i forgot about some of the reasoning behind this, so i had to review the code again. This pull request is correct, and basically the environment variables i remove are set in code [here](https://github.com/chef/chef/blob/3ad740754507629a73525ffdcc0b3eb06f177001/lib/chef/application.rb#L178) the config options for the aforementioned code snippet are set in the pr instead.

This is therefore correct.

tgagor:
Those ENV variables are required not for chef-client but for other mechanisms like:
* gem install ...
* npm install ...
* git clone http://...

Chef-client will run OK but other commands will crash (and this is exactly what I'm expecting after cookbook update from 4.0.0 to 4.3.1

Please restore those variables additionally to standard http(s)_proxy options.

willejs:
In the link I reference to above, if the env vars are not set the code sets them? Then executes the chef run.

tgagor:
Somehow this doesn't work, until I configure ENV variables in client.rb in the old way.

juliandunn:
@tgagor I don't think we're going to restore these. The cookbook is overloaded in functionality already and configuring general shell environment proxy parameters for the system is clearly out of scope.

There is a way for you to configure additional parameters not accounted for in this cookbook: the `/etc/chef/client.d` directory into which you can drop additional files, templates, etc. from your own cookbook.

tgagor:
Thanks, I've done this like that.
If you're not going this way you should consider removing proxy trickery from main client.rb, ex:
https://github.com/chef-cookbooks/chef-client/blob/master/templates/default/client.rb.erb#L32
Standard config generator will do this properly and this look like abandoned after removal of variables.

PR #241, opened 2015-01-11
Configure logrotate to use systemd when appropriate in chef-client::config recipe.

jhh:
This fixes a bug on Centos 7 where logrotate was trying to use a nonexistant `/etc/init.d/chef-client` script on a `systemd` system.

### DIFF
--- a/recipes/config.rb
+++ b/recipes/config.rb
@@ -38,11 +38,12 @@ if node['chef_client']['log_file'].is_a? String and node['chef_client']['init_st
       rotate node['chef_client']['logrotate']['rotate']
       frequency node['chef_client']['logrotate']['frequency']
       options ['compress']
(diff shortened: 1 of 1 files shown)


jtimberman:
I think using the implicit return from a conditional would be easier to understand and extend in the future. This uses a `case` statement instead:

```ruby
  logrotate_app 'chef-client' do
    path [log_path]
    rotate node['chef_client']['logrotate']['rotate']
    frequency node['chef_client']['logrotate']['frequency']
    options ['compress']
    postrotate case node['chef_client']['init_style']
               when 'systemd'
                 'systemctl reload chef-client.service >/dev/null || :'
               else
                 '/etc/init.d/chef-client reload >/dev/null || :'
               end
  end
```

jhh:
Agree and thanks for the suggestion, updated the pull request.

PR #246, opened 2015-01-16
add runit finish option

ptqa:
If someone else needs finish script support here.

### DIFF
--- a/metadata.rb
+++ b/metadata.rb
@@ -4,7 +4,7 @@ maintainer_email  'cookbooks@opscode.com'
 license           'Apache 2.0'
 description       'Manages client.rb configuration and chef-client service'
 long_description  IO.read(File.join(File.dirname(__FILE__), 'README.md'))
-version           '4.0.0'
+version           '4.0.1'
 recipe 'chef-client', 'Includes the service recipe by default.'
 recipe 'chef-client::arch_service', 'Configures chef-client as a service on Arch Linux'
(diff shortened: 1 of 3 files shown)


metadata.rb
```diff
@@ -4,7 +4,7 @@
 license           'Apache 2.0'
 description       'Manages client.rb configuration and chef-client service'
 long_description  IO.read(File.join(File.dirname(__FILE__), 'README.md'))
-version           '4.0.0'
+version           '4.0.1'
```
jtimberman: Please don't bump the version in pull requests. The maintainers will handle versioning.

jtimberman:
Aside from the note about versions what is the purpose of this? What's the use case?

ptqa:
@jtimberman with this you can override this attribute and use finish script for removing temporary files, killing hanged up processes, waiting for 5 sec, etc.

jtimberman:
I understand the purpose of a finish script with runit. What is the use case with chef-client? I'm hesitant to have yet another knob to turn (attributive) that can increase the cognitive load for someone trying to implement this cookbook in their environment. In 6 years of running chef-client in runit I've not personally needed to have a finish script - the client behaves pretty well for me. 

Is there some behavior of chef-client that needs to have clean up performed like this? Perhaps there is a bug we should address in chef itself?

ptqa:
@jtimberman well, the problem I faced is chef-client hangs up, but I have not yet managed to find out why. So I run chef-client using timeout util and use finish script for waiting interval time. I understand that it's not a best way to manage it.

Issue #196, opened 2014-09-09
Chef server 12 runs under 'opscode' user, not 'chef_server'


dwradcliffe:
This `dir_owner` helper will return 'chef_server' and thus the config recipe will fail every time.

```
Chef::Exceptions::UserIDNotFound: template[/etc/chef/client.rb] (chef-client::config line 73) had an error: Chef::Exceptions::UserIDNotFound: cannot determine user id for 'chef_server', does the user exist on this system?
```

https://github.com/opscode-cookbooks/chef-client/blob/bf1dbbc749c92315784b17bac796ed858a256f5d/recipes/config.rb#L70-L85

https://github.com/opscode-cookbooks/chef-client/blob/bf1dbbc749c92315784b17bac796ed858a256f5d/libraries/helpers.rb#L60-L66

https://github.com/opscode-cookbooks/chef-client/blob/bf1dbbc749c92315784b17bac796ed858a256f5d/libraries/helpers.rb#L26-L28

gplessis:
Got bitten by this annoying issue today, after a Chef Server upgrade from 11 to 12. @jtimberman, @binamov, any update/status on this?

joshskains:
Is there a solution to this? After a brutal upgrade from 11 to 12, we now have a chef-client break on our main chef server. That's really unacceptable. How did this get through QA?

BarthV:
:+1: 
Didn't find out a workaround for this, pls make a move :)

gplessis:
FYI using [chef-rewind](https://github.com/thommay/chef-rewind) is one possible temporary workaround to make `chef-client::config` converge on a system with Chef Server 12 installed.
```ruby
chef_gem "chef-rewind"
require 'chef/rewind'

include_recipe 'chef-client::config'
rewind :template => "#{node["chef_client"]["conf_dir"]}/client.rb" do
  owner 'root'
  group 'root'
end
```

caquino:
Same issue here.

   * cannot determine user id for 'chef_server', does the user exist on this system?
    ================================================================================
    Error executing action `create` on resource 'template[/etc/chef/client.rb]'
    ================================================================================

    Chef::Exceptions::UserIDNotFound
    --------------------------------
    cannot determine user id for 'chef_server', does the user exist on this system?

justindossey:
I worked around this with
```
include_recipe 'chef-client::config'
r = resources(template: '/etc/chef/client.rb')
r.user 'root'
r.group 'root'
```

[...]

PR #259, opened 2015-02-04
Removed dir_group/dir_owner

BackSlasher:
As per discussion in #196, created PR without a special case to Chef Server
Did not remove `chef_server?` as it's still being used.

For your consideration

### DIFF
--- a/libraries/helpers.rb
+++ b/libraries/helpers.rb
@@ -57,14 +57,6 @@ module Opscode
         end
       end
 
-      def dir_owner
-        if chef_server?
-          chef_server_user
-        else
-          root_owner
-        end
-      end
-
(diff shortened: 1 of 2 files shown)


dwradcliffe:
Assuming there's no special reason for this, looks good to me.

r: @jtimberman

juliandunn:
In PR #253 I'm in the middle of excising all the Chef Server-specific logic from this cookbook -- I just need to test the oldest supported Enterprise Chef 11 to make sure it doesn't throw things in `/etc/chef`.

jtimberman:
I'm in favor of #253, removing the `chef_server` detection cruft would help make this all more maintainable and understandable.

PR #269, opened 2015-02-17
Do not duplicate proxy configuration

cedric-lamalle:
Added http_proxy, https_proxy and no_proxy options to the ignore list, else they get duplicated in #{node["chef_client"]["conf_dir"]}/client.rb.
See lines 31-40.

### DIFF
--- a/templates/default/client.rb.erb
+++ b/templates/default/client.rb.erb
@@ -9,7 +9,7 @@ end
 
 <% end -%>
 <% @chef_config.keys.sort.each do |option| -%>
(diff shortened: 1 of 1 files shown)


dcparker88:
This looks like it might have gotten removed. I am seeing this same behavior, and those values are no longer present in https://github.com/opscode-cookbooks/chef-client/blob/master/templates/default/client.rb.erb#L12. Any reason they are removed now?

cedric-lamalle:
Commit 0fad4c6 reverted commit a6a8735 which was the commit just before 30e948e. I think this has been removed by mistake.
I can make a new pull request.

dcparker88:
Thank you!

PR #257, opened 2015-02-02
Add file permission attribute to /var/log/chef

jsmickey:
I have a use case where my team needs to change the permissions of /var/log/chef/ for a third party log agent.  I added an attribute for the file permissions and set the default to 750.  This change will allow for changing the permissions by overriding the attribute.

PR #274, opened 2015-02-25
Enable changing of permissions for log directory

itwasntandy:
Previously the permissions for `node['chef_client']['log_dir']` were
hardcoded to 00750.
Sometimes it can be helpful for non-root users to view Chef logs,
and at othertimes, users may wish to set log_dir to /var/log.

In both cases it can be useful to allow the easy changing of mode
for this directory. This change permits it via an attribute.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,7 @@ The following attributes affect the behavior of the chef-client program when run
 * `node['chef_client']['log_dir']` - Sets directory used in
   `Chef::Config[:log_location]` via command-line option to a location
   where chef-client should log output. Default "/var/log/chef".
+* `node['chef_client']['log_dir_mode']` - Sets permissions for the log directory. Default 00750.
 * `node['chef_client']['log_rotation']['options']` - Set options to logrotation of chef-client log file. Default `['compress']`.
 * `node['chef_client']['log_rotation']['postrotate']` - Set postrotate action for chef-client logrotation. Default to chef-client service reload depending on init system.
 * `node['chef_client']['conf_dir']` - Sets directory used via
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -47,6 +47,7 @@ default['chef_client']['bin']         = '/usr/bin/chef-client'
 # Set a sane default log directory location, overriden by specific
 # platforms below.
 default['chef_client']['log_dir']     = '/var/log/chef'
+default['chef_client']['log_dir_mode'] = 00750
 
 # Configuration for chef-client::cron recipe.
 default['chef_client']['cron'] = {
--- a/libraries/helpers.rb
+++ b/libraries/helpers.rb
@@ -96,7 +96,7 @@ module Opscode
           rescue Chef::Exceptions::ResourceNotFound
             directory node['chef_client'][dir] do
               recursive true
-              mode 00750 if dir == 'log_dir'
+              mode node['chef_client']['log_dir_mode'] if dir == 'log_dir'
               owner d_owner
               group d_group
             end


robcoward:
:+1:

jtimberman:
I'm still :-1: on this idea, as I was here https://github.com/opscode-cookbooks/chef-client/pull/257

Creating attributes for this kind of thing, to make every single bit customizable defeats the purpose of a general cookbook that fits the broadest general use case. It also increases the complexity of maintenance and testing because we have two conditions to test against.

itwasntandy:
I take your point with regard to complexity and maintenance. However equally burying hardcoded special case defaults cause pain and disruption..

If the goal was to be generic, and simple, why have the special case for the log_dir directory, in setting the mode to 0750 ?
``` 
mode 00750 if dir == 'log_dir'
```
Why not just rely on what has been configured for OS defaults in that case? - as that would truly leave the end result up to the user.

Hardcoding permissions like this can be disruptive, as it's somewhat non obvious to a user of such a cookbook.

We hit this because we wanted to enable log rotation for Chef logs, and run chef from cron.
Previously `node['chef_client']['cron']['log_file]` was set to /var/log/chef-client.log

The log rotation stuff in the chef client cookbook relies on the values of  `node['chef_client']['log_dir']` and `node['chef_client']['log_file'] so we set those appropriately (i.e. log_dir to /var/log, and log_file to chef_client.log) and altered the cron log_file to be a File.join of those attributes.

The net result was /var/log being set to 00750, which was somewhat disruptive.

Of course _now_ we have test cases in place of the permissions of /var/log but the lack of documented clarity regarding the effect of setting `node['chef_client']['log_dir']` to an existing directory caused avoidable disruption.

jtimberman:
> If the goal was to be generic, and simple, why have the special case for the log_dir directory, in setting the mode to 0750 ?

That particular line has a fairly sordid history which took some digging to find. Logic around the log directory mode specifically was added in dde507750523ee86332b45edad23b6461a810d57 which is where it appears the permissions were changed to remove "other" read/execute permission per [COOK-2052](https://tickets.opscode.com/browse/COOK-2052) as an "insecure default." If anything, I'd argue against the log dir original mode of `00755` being an insecure default, and *that* it should be the default hard coded value, effectively reverting that change.

My point about being simple is not about what we've done in the past, but how we maintain this in the future. This cookbook is very, clearly, absolutely, **not** generic and simple. It has cruft and technical debt that makes it, in my opinion, difficult to maintain going forward. It tries to do too much and provide too many knobs for the user to customize because it accommodates so many use cases and bikesheds. Adding more attributes increases complexity. More attributes means more testing to guard against regressions.

> The log rotation stuff in the chef client cookbook relies on the values of node['chef_client']['log_dir'] and node['chef_client']['log_file'] so we set those appropriately (i.e. log_dir to '/var/log', and log_file to 'chef_client.log') and altered the cron log_file to be a File.join of those attributes.

IMO we shouldn't have a `log_dir` attribute. We should have only `log_file` and have a helper method that can return the log directory using `File.dirname`. At the very least, the README should be clear that this directory should be specifically for Chef, and not a general purpose directory. That's a common convention with non-system daemons, e.g. `/var/log/apache2` or similar.

To summarize the above, rather than adding another attribute, I prefer that we:

1. Clarify the documentation about the `node['chef_client']['log_dir']` attribute that it is for Chef's logs, and should not be set to a shared log location.
2. Set the mode to `00755` as it was before commit dde507750523ee86332b45edad23b6461a810d57.

itwasntandy:
> To summarize the above, rather than adding another attribute, I prefer that we:

> 1. Clarify the documentation about the `node['chef_client']['log_dir']` attribute that it is for Chef's logs, and >     should not be set to a shared log location.
> 2. Set the mode to `00755` as it was before commit [dde5077](https://github.com/opscode-cookbooks/chef-client/commit/dde507750523ee86332b45edad23b6461a810d57).

+1 on that. That sounds like a good approach. Would you like me to prepare a pull request for that? - or shall I update this one with that?

Thank you also for explaining a little on the future direction of the cookbook. I've noticed various inconsistencies in there, and will be happy to help tidy and simplify it in future.

jtimberman:
Make a new PR.

jtimberman:
And thanks :cake:

chef-supermarket:
Hi. I am an automated pull request bot named Curry. There are commits in this pull request whose authors are not yet authorized to contribute to Chef Software, Inc. projects or are using a non-GitHub verified email address. To become authorized to contribute, you will need to sign the Contributor License Agreement (CLA) as an individual or on behalf of your company. [You can read more on Chef's blog.](https://www.chef.io/blog/2014/06/23/changes-to-the-contributor-license-agreement-process)

juliandunn:
We've settled on just changing the log directory's permissions `0755` in #257 so I'm closing this now.

PR #275, opened 2015-02-26
correct logrotate postrotate command for upstart init_style

dupuy:
If init_style is set explicitly to 'upstart' the /etc/init.d/chef-client script will not exist, use initctl instead


dupuy:
I think this is covered by the "obvious fix" clause, but let me know if you need me to sign a CLA.

juliandunn:
:+1:

PR #277, opened 2015-02-26
Remove outdated documentation, replace with links

dupuy:
Previous content was outdated and incorrect, and had several broken links.
Rather than copy or create updated information (which will become outdated), replace with links to documentation and CONTRIBUTING.md for Chef itself (assuming that with two unrelated links we increase the likelihood that one of them will still work in the future).

### DIFF
--- a/CONTRIBUTING
+++ b/CONTRIBUTING
@@ -1,29 +1,5 @@
-If you would like to contribute, please open a ticket in JIRA:
+If you would like to contribute, please see
 
-* http://tickets.chef.io
+* https://docs.chef.io/community_contributions.html
 
-Create the ticket in the COOK project and use the cookbook name as the
-component.
-
-For all code contributions, we ask that contributors sign a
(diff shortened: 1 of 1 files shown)


jtimberman:
Having updated documentation is good. However:

1. Cookbooks aren't maintained in the same way as Chef itself. Much of the Chef contribution document doesn't apply to cookbooks for various reasons.
2. We need to update all the cookbooks we maintain not just this one, and we need a consistent document to use. I don't know that Chef's CONTRIBUTING.md is the right thing.

Thoughts, @someara @nathenharvey?

jtimberman:
(that's not to say that perhaps the contributing document in the chef repository couldn't be adapted to accommodate cookbooks :smile:)

juliandunn:
Let's just take this for now -- perfect is the enemy of "shipped", and the docs are currently totally wrong.

PR #279, opened 2015-03-03
Updates for Mac OS X support

danieljimenez:
1. Rename plist to proper name (broken in master)
2. Add support for node['chef_client']['task']['user'].

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -173,6 +173,8 @@ default['chef_client']['log_rotation']['options'] = ['compress']
 default['chef_client']['log_rotation']['postrotate'] =  case node['chef_client']['init_style']
                                                         when 'systemd'
                                                           'systemctl reload chef-client.service >/dev/null || :'
+                                                        when 'upstart'
+                                                          'initctl reload chef-client >/dev/null || :'
                                                         else
                                                           '/etc/init.d/chef-client reload >/dev/null || :'
                                                         end


.idea/misc.xml
```diff
@@ -0,0 +1,201 @@
...
+              <list>
+                <option value="0.2" />
+              </list>
+            </option>
+          </splitter-proportions>
+        </settings>
+      </state>
+    </states>
+  </component>
+</project>
```
jtimberman: Please remove this file.

metadata.rb
```diff
@@ -4,7 +4,7 @@
 license           'Apache 2.0'
 description       'Manages client.rb configuration and chef-client service'
 long_description  IO.read(File.join(File.dirname(__FILE__), 'README.md'))
-version           '4.2.4'
+version           '4.2.41'
```
jtimberman: Please don't modify the version in metadata.

recipes/launchd_service.rb
```diff
@@ -15,19 +15,38 @@ class ::Chef::Recipe
...
 
 if mac_service_supported
-  template '/Library/LaunchDaemons/com.chef.chef-client.plist' do
-    source 'com.chef.chef-client.plist.erb'
-    mode 0644
-    variables(
-      :launchd_mode => node['chef_client']['launchd_mode'],
-      :client_bin => client_bin
-    )
+  if node['chef_client']['task']['user'].eql? 'SYSTEM'
```
jtimberman: The `node['chef_client']['task']['user']` attribute is for Windows, specifically for managing the Chef Client as a scheduled task (Windows' cron-alike).

attributes/default.rb
```diff
@@ -137,6 +137,7 @@
   # as 'interval' if you want chef-client to be run
   # periodically by launchd
   default['chef_client']['launchd_mode'] = 'interval'
+  default['chef_client']['launchd_user'] = 'root'
```
jtimberman: I think we can just hardcode `'root'` in the recipe, or use the `dir_owner` and `dir_group` helper methods in the cookbook.

.idea/misc.xml
danieljimenez: Ahh sorry, will do.

recipes/launchd_service.rb
danieljimenez: Any suggestions as to what to use?

recipes/launchd_service.rb
```diff
@@ -15,19 +15,38 @@ class ::Chef::Recipe
...
+      variables(
+          :launchd_mode => node['chef_client']['launchd_mode'],
+          :client_bin => client_bin
+      )
+    end
+  else
+    home_dir = Dir.home(node['chef_client']['task']['user'])
+    
+    directory "#{home_dir}/Library/LaunchDaemons" do
+      owner node['chef_client']['task']['user']
```
jtimberman: If the intention here is to allow the `chef-client` service to run as a non-root user, is it worth supporting that use case? Some operations that one would want to do with Chef will require privileged access, or at least the use of sudo to provide privileged access. That won't work when run daemonized, since it will sit and wait for a password prompt...

danieljimenez: Have you seen the pivotal-sprout repos? They require use of sudo and will not run as root.

PR #286, opened 2015-03-11
Remove legacy chef-server handling

jtimberman:
This commit merges #253, which resolves the following issues or PR's:

- Closes #196
- Closes #203
- Closes #259

It assumes Chef Server 12 for testing, but should work on systems that are running Chef Server 11 or Enterprise/Private Chef as those products are self-contained and don't have files in the directories managed by the recipes in question in this cookbook (namely config, but handled through the helper library to create directories).

I'd like to know if this works for the folks who have been involved in
those various issues/pull requests, so mentioning here:

@BackSlasher, @dwradcliffe, @juliandunn, @afiune, @binamov, @JeanMertz, @cwebberOps, @hatchetation, @jblaine, @BarthV, @gplessis @caquino, @justindossey, @ikurochkin

PR #287, opened 2015-03-11
Introduce 'dir_owner' and 'dir_group' attributes

ymattw:
Recipe chef-client::config tries to set file ownership to user 'chef_server' returned by helper method 'chef_server_user' on Chef 12, which does not exist on our Chef servers. Talked to a support engineer, he mentioned this could be a edge-case bug. This change is to give user flexibility to customize the owner and group.


jtimberman:
I believe this is unnecessary with #286.

jtimberman:
(And I plan to merge that later this week, wanted to give folks who commented on the various issues/pull-requests a chance to test that it works in their environments).

PR #289, opened 2015-03-14
Set init style to systemd for Debian 8 systems if init_package == systemd

ve7cxz:
Fresh installs of Debian 8.0 (jessie) now use systemd by default. As it stands, the chef-client cookbook assumes an init style of 'init' for Debian servers, which will fail for Debian 8.0.

These changes add a switch to the attributes to set the init style to 'systemd' if the platform_version is >= 8.0, and if init_package is set to 'systemd'.


jtimberman:
This is probably fine. I don't like all the version and platform conditionals in this cookbook but it's the mess we made. Is Jessie released? Or does it have platform version populated correctly like that? 

We should rely on init_package for duck typing this wrt https://github.com/opscode-cookbooks/chef-client/issues/282

ve7cxz:
Yeah, editing the attributes made me cry a bit inside but it does work.

As to whether it's released yet - I don't believe so, but this was a fresh install from a jessie ISO image.

PR #295, opened 2015-03-19
Add support for specifying gem source for config.

johnbellone:
For us poor folks that are behind the corporate veil we need ability to
not hit a gem server for core gems such as chef-vault, chef-reporting
and chef-sugar.


jtimberman:
Seems legit.

PR #308, opened 2015-04-15
Fix delete_validation.rb exception bug when node['chef-client']['config']['validation_key']` is nil

ozzyogkush:
Fixes a bug where provisioning with chef_zero results in a nil value for the `node['chef-client']['config']['validation_key']` attribute, causing the file resource to be looking at a blank name and hence raising an exception.

See https://github.com/opscode-cookbooks/chef-client/issues/307

### DIFF
--- a/recipes/delete_validation.rb
+++ b/recipes/delete_validation.rb
@@ -25,9 +25,6 @@ unless Chef::Config[:validation_key].nil?
   file Chef::Config[:validation_key] do
     action :delete
     backup false
-    only_if {
-      ::File.exists?(Chef::Config[:validation_key]) &&
-      ::File.exists?(Chef::Config[:client_key])
-    }
+    only_if { ::File.exists?(Chef::Config[:client_key]) }
   end
 end
--- a/recipes/delete_validation.rb
+++ b/recipes/delete_validation.rb
@@ -21,8 +21,13 @@ class ::Chef::Recipe
   include ::Opscode::ChefClient::Helpers
 end
 
-file Chef::Config[:validation_key] do
-  action :delete
-  backup false
-  only_if { ::File.exists?(Chef::Config[:client_key]) }
+unless Chef::Config[:validation_key].nil?
+  file Chef::Config[:validation_key] do
+    action :delete
+    backup false
+    only_if {
+      ::File.exists?(Chef::Config[:validation_key]) &&
+      ::File.exists?(Chef::Config[:client_key])
+    }
+  end
 end


chef-supermarket:
Hi. I am an automated pull request bot named Curry. There are commits in this pull request whose authors are not yet authorized to contribute to Chef Software, Inc. projects or are using a non-GitHub verified email address. To become authorized to contribute, you will need to sign the Contributor License Agreement (CLA) as an individual or on behalf of your company. [You can read more on Chef's blog.](https://www.chef.io/blog/2014/06/23/changes-to-the-contributor-license-agreement-process)

ozzyogkush:
I did sign the ICLA prior to pulling this.

juliandunn:
How does `Chef::Config[:validation_key]` even become `nil`? That's what I don't understand -- it has a default value internal to Chef.

recipes/delete_validation.rb
```diff
@@ -21,8 +21,13 @@ class ::Chef::Recipe
...
-file Chef::Config[:validation_key] do
-  action :delete
-  backup false
-  only_if { ::File.exists?(Chef::Config[:client_key]) }
+unless Chef::Config[:validation_key].nil?
+  file Chef::Config[:validation_key] do
+    action :delete
+    backup false
+    only_if {
+      ::File.exists?(Chef::Config[:validation_key]) &&
```
juliandunn: You don't actually need to add this extra guard, since we want this resource to fire <i>iff</i> a client key exists.

ozzyogkush:
Chef zero fakes the validation process, though I'm not clear on the internals of that. Perhaps at some point that config var gets unset. I'm using the latest ChefDK version.

chef-supermarket:
Hi. Your friendly Curry bot here. Just letting you know that all commit authors have become authorized to contribute. I have added the "Signed CLA" label to this issue so it can easily be found in the future.

recipes/delete_validation.rb
ozzyogkush: I've reverted the guard.

juliandunn:
Ah, I figured it out -- my understanding was from a pre-Chef-11.8.0 world, when the behavior changed: https://github.com/chef/chef/commit/bbea73b15e3a20b344027bf454e31ce1c31fbc6a#diff-ce8cfba0deb02929653a78f574664565R310

So this PR makes sense -- if you can just remove the extra guard then we can merge.

ozzyogkush:
Latest commit has reverted the guard change. Thanks for the quick feedback.

juliandunn:
Squashed and merged in 4d4dd0db860a2380e163f19a9c8a2d6371f090f8

PR #319, opened 2015-07-07
update syslog to use new shiny 12.4 helper

josephholsten:
also `s/include/include_recipe`, and removing hat tip.

oh, I considered adding a bit for:
```ruby
default_attributes(
  "chef_client" => {
    "config" => {
      "log_level" => false
    }
  }
)
```
to remove it from the main `client.rb`. Not strictly necessary, of course.


jtimberman:
:heart:

Issue #235, opened 2014-12-21
Setting node["chef_client"]["interval"] seems to be ignored for Windows service


adamlau:
I installed chef-client 12.0.3 via bootstrap onto a Windows 8.1 machine. I used the default chef-client recipe to install the windows chef-client service. I'm trying to set the interval for the service to run to 5 mins.

I have a base role set for the windows node:

    name "base"
    description "Base role"
    run_list "recipe[chef-client]"
    default_attributes 
    override_attributes "chef_client" => { "interval" => "300" }

The chef-client windows service gets installed correctly but it runs at the default of 1800 seconds. I've tried a few different ways of setting the attribute but nothing seems to budge it from the default.

The documentation (https://github.com/opscode-cookbooks/chef-client#attributes) says:
> The following attributes affect the behavior of the chef-client program when running as a service through one of the service recipes, or in cron with the cron recipe, or are used in the recipes for various settings that require flexibility.
>
> * node["chef_client"]["interval"] - Sets Chef::Config[:interval] via command-line option for number of seconds between chef-client daemon runs. Default 1800.

I've looked through the source code of the chef-client recipe and executable but can't find how the interval is set via command-line option when using the windows service recipe. Do I have to set the interval attribute myself in client.rb?

jfbibeau:
Having the same issue on Windows 7...

I have the following in my wrapper cookbook that invokes the chef-client recipe:

```
default['chef_client']['interval'] = '172800'
default['chef_client']['splay'] = '3600'
```

Works on Linux fine, but on Windows the service is stuck to 1800 seconds. I can confirm because at the end of the converge, the service prints:

```
[2015-01-14T08:35:12-08:00] INFO: Next chef-client run will happen in 1800.3427924620066 seconds
```

echohack:
+1 here. This is effecting all Windows boxes in our infrastructure, including Windows 2008R2, W7, and 2012R2.

Setting a policy that ALL nodes can follow is super important for us so we can determine the check-in behavior of the nodes.

juliandunn:
It's because the current command to set up a Windows service predates the ability of the service manager to handle `-i` and `-s` flags for interval and splay respectively:

https://github.com/opscode-cookbooks/chef-client/blob/master/recipes/windows_service.rb#L33

I would take a PR to add them and populate using the attributes.

echohack:
[...]

PR #322, opened 2015-07-09
Support interval and splay on Windows

jaym:
This is the simplest fix I can think of for #235 while not making chef/chef#3432 an issue. We can also make better guarantees around idempotence if we're just modifying a file (the service manager currently only checks if the service is installed, which is quite incorrect as you could modify the interval in your cookbook and nothing would change) This fix involves no changes to chef client so it will work with any versions of chef this cookbook already works with. One issue I did find with chef client was that the service manager appends -L with a default value if you don't specify it, which means we cannot put the log location in `client.service.rb`.

cc @jeremiahsnapp @smurawski @juliandunn

### DIFF
--- a/recipes/windows_service.rb
+++ b/recipes/windows_service.rb
@@ -29,8 +29,19 @@ end
 # libraries/helpers.rb method to DRY directory creation resources
 create_directories
 
+d_owner = root_owner
+d_group = node['root_group']
+
+template "#{node["chef_client"]["conf_dir"]}/client.service.rb" do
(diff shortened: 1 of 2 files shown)


smurawski:
:+1:

jeremiahsnapp:
:+1: thanks @jaym

ksubrama:
:+1:

siwy202:
:+1:

juliandunn:
I'm shipping it. Thanks @jaym

PR #327, opened 2015-08-07
Fix chef windows task quoting

Annih:
Former task command was using single quotes, relying on windows_task
implementation - based on schtasks.exe - that replace them by double
quotes.
A different windows_task implementation might change this behavior.


jaym:
sounds reasonable :+1:

PR #318, opened 2015-06-25
Fix windows_task configuration

smurawski:

PR #329, opened 2015-08-11
added check for nil attributes for windows_task creds

### LINKED ISSUE #178: Add user and password attribute on the recipe task for windows.
I installed the recipe task.rb from the cookbook chef-client for Windows and I noticed that the "run whether user is logged on or not" is not enabled, it seems when you want this option enabled in the windows_task resource you need add the user and password which this task will run, so is a good idea to add this option on the task.rb and also would be a good idea where you can get the password via data_bag.

burnzy:
Since windows_task will by default use the SYSTEM account, we can change the attributes to nil.
This will avoid having to specify a password for the SYSTEM account.

I believe this fixes #178


burnzy:
closing this, was already addressed in #318

PR #331, opened 2015-08-17
Fix client bin in dummy starter script when daemon options are set

jf647:
Clone and modify a copy of client_bin in the cron recipe so that the
dummy daemon starter script just has the path to chef-client but
the cronjob has the path plus any daemon options that are set.

This fixes the use case where chef-client is invoked from cron but
node['chef_client']['daemon_options] is set.  Prior to this change,
setting daemon_options would result in the dummy starter script
being written like this:

```
DAEMON=/usr/bin/chef-client --option value
```

which would then fail the "test -x $DAEMON" line later in the starter.

The daemon starter gets called with 'reload' from the logrotate script,
even if you are invoking chef-client from cron.

### DIFF
--- a/recipes/cron.rb
+++ b/recipes/cron.rb
@@ -79,8 +79,15 @@ env        = node['chef_client']['cron']['environment_variables']
 log_file   = node['chef_client']['cron']['log_file']
 append_log = node['chef_client']['cron']['append_log'] ? '>>' : '>'
 
-# Use daemon_options in cron.
-client_bin << " #{node["chef_client"]["daemon_options"].join(' ')}" if node["chef_client"]["daemon_options"].any?
+# Use daemon_options in cron, but don't override client_bin because
+# then we'll create a dummy service starter that looks something like
+#
+# DAEMON=/usr/bin/chef-client --option value
+#
+# which then fails to pass 'test -x $DAEMON'
+cron_client_bin = client_bin.dup
+cron_client_bin << " #{node["chef_client"]["daemon_options"].join(' ')}" \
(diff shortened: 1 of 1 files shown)


recipes/cron.rb
```diff
@@ -98,7 +105,7 @@ class ::Chef::Recipe
     user    'root'
     cmd = ''
     cmd << "/bin/sleep #{sleep_time}; " if sleep_time
-    cmd << "#{env} #{client_bin} #{append_log} #{log_file} 2>&1"
+    cmd << "#{env} #{cron_client_bin} #{append_log} #{log_file} 2>&1"
```
juliandunn: This solution seems super messy. Why can't we just do something like

```ruby
daemon_options = " #{node["chef_client"]["daemon_options"].join(' ')}" if node["chef_client"]["daemon_options"].any?
...
cmd << "#{env} #{client_bin} #{daemon_options} #{append_log} #{log_file} 2>&1"
```

instead of all the `dup` stuff?

chef-supermarket:
Hi. I am an automated pull request bot named Curry. There are commits in this pull request whose authors are not yet authorized to contribute to Chef Software, Inc. projects or are using a non-GitHub verified email address. To become authorized to contribute, you will need to sign the Contributor License Agreement (CLA) as an individual or on behalf of your company. [You can read more on Chef's blog.](https://www.getchef.com/blog/2014/06/23/changes-to-the-contributor-license-agreement-process)

tas50:
I'm going to close this out since we haven't heard back from the author after our review.

PR #337, opened 2015-08-26
Updates .travis.yml to use the travis task

### LINKED ISSUE #336: rake spec vs rake travis in travis.yml
I see we have a nice `rake travis` task in the `Rakefile`. Is there a reason we instead call the `rake spec` in the `travis.yml`?

BTW, I never would have noticed had not the build on _master_ been failing.

philoserf:
Use the _travis_ task rather than the generic _spec_ task. Closes #336

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,4 +5,4 @@ rvm:
 - 2.1.5
 - 2.2.0
 script:
-- rake spec
+- rake travis


philoserf:
Now that is what happens when we don't run our tests/enforce our standards. Foodcritic style check failures in code that has already gotten past the censer. Shall we fix FC034 in the file, suppress it, adjust in some other way?

tas50:
We've changed this cookbook to use ChefDK in travis so this change is no longer applicable

PR #343, opened 2015-09-29
fixes for freebsd cron invocation and kitchen changes required to test

lamont:
the cron recipe would not converge under freebsd, there's no init.d equiv and I thought the cleanest way would be to put a pair of not_if guards on the two unneeded templates. I also had to add the '"/bin/sh"' to the .kitchen.yml to get kitchen to work at all with the freebsd box files. I tested that the freebsd/cron suite works under vagrant and that my '"/bin/sh"' changes do not break at least the ubuntu suites.

### DIFF
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -3,8 +3,6 @@ driver:
   customize:
     memory: 1024
     cpus: 2
-  ssh:
-    shell: '"/bin/sh"'
 
 provisioner:
   name: chef_zero
@@ -18,6 +16,9 @@ platforms:
   - name: fedora-21
   - name: freebsd-9.3
   - name: freebsd-10.1
+    driver:
+      ssh:
+        shell: '/bin/sh'
   - name: ubuntu-12.04
   - name: ubuntu-14.04
 
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -3,6 +3,8 @@ driver:
   customize:
     memory: 1024
     cpus: 2
+  ssh:
+    shell: '"/bin/sh"'
 
 provisioner:
   name: chef_zero
--- a/recipes/cron.rb
+++ b/recipes/cron.rb
@@ -41,18 +41,16 @@ dist_dir, conf_dir = value_for_platform_family(
 
 # let's create the service file so the :disable action doesn't fail
 case node['platform_family']
-when 'arch', 'debian', 'rhel', 'fedora', 'suse', 'openbsd', 'freebsd'
+when 'arch', 'debian', 'rhel', 'fedora', 'suse', 'openbsd'
   template '/etc/init.d/chef-client' do
     source "#{dist_dir}/init.d/chef-client.erb"
     mode 0755
     variables(client_bin: client_bin)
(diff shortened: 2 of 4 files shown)


lamont:
I note that travis is and has been failing but PR #337 fixes that issue. My local kitchen tests succeeded, but it would be nice to get the green travis badge of happiness as well.

lamont-granquist:
lol, confusing...

lamont:
let's combine our lamont powers for great justice!  But on a more serious note, I'm happy to yank out my .travis.yml changes if the other PR does a better job. the .kitchen.yml change was required to get freebsd working under kitchen.

recipes/cron.rb
```diff
@@ -46,11 +46,13 @@ class ::Chef::Recipe
     source "#{dist_dir}/init.d/chef-client.erb"
     mode 0755
     variables(client_bin: client_bin)
+    not_if { node['platform_family'] == 'freebsd'}
```
lamont-granquist: that should probably not be a not_if and should wrap the template completely so it doesn't even appear as a resource on freebsd.

foodcritic might get annoyed by this, but foodcritic can kinda suck it and you can use `# ~FCXYZ` to turn it off.

.kitchen.yml
```diff
@@ -3,6 +3,8 @@ driver:
   customize:
     memory: 1024
     cpus: 2
+  ssh:
+    shell: '"/bin/sh"'
```
lamont-granquist: doesn't TK take driver arguments under the platform section as well, so that we can set this only for freebsd?

lamont-granquist:
I don't have a lot of opinion on `rake spec` vs `rake travis`, it should probably `bundle exec` either way for less confusion.

lamont:
Ok, I'll rework the .kitchen.yml change so that the ssh/shell is applied to the freebsd platform specifically and put the two templates in an anti-freebsd block then ping you again. Probably Thursday or Friday.  And I'll add the .foodcritic ignore for the warning I'm introducing.

lamont:
Well, that fixes rake spec on my machine, but the TravisCI build fails cause we hit the 4 megabyte log limit due to the repeated warnings about needing override: true in new resources. 

WARN: You declared a new resource #<Class:0x007fbd95c3d430> for resource windows_certificate, but it comes alphabetically after LWRP resource windows_certificate from cookbook windows and has the same filters ({}), so it will not be used. Use override: true if you want to use it for windows_certificate.

I have no idea why that warning suddenly appeared, but it's affected other tests I was running using the apt cookbook. Happened under chefDK 0.7.0 and also 0.8.1.

josephholsten:
@lamont did this ever get complete to your satisfaction?

lamont-granquist:
looks fine to me but looks like it needs a serious rebase.

dheerajd-msys:
@lamont Please check ^

tas50:
I'm going to close this one out at this point due to the age and drift in the codebase. If this is still something you want to do let me know in the community slack and we can figure out next steps.

PR #347, opened 2015-11-04
add windows service log location to client.service.rb and only pass to service manager for legacy clients

mwrock:
This PR depends on https://github.com/chef/chef/pull/4135

This will add the `log_location` to the client.service.rb and as long as the chef client version is greater than 12.5.1, it will not pass the `log_location` to the windows-service binary which will get the location from the client.service.config.

This will allow users to change the log location in the config whereas now the path is immutable after the service is installed.

### DIFF
--- a/libraries/helpers.rb
+++ b/libraries/helpers.rb
@@ -31,7 +31,7 @@ module Opscode
       end
 
       def chef_client_service_running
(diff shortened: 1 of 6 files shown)


chefsalim:
:+1:

ksubrama:
:+1:

btm:
:+1: you should probably also ship this, and then send an email to the mailing list explaining how the defaults are going to change so google indexes those words.

PR #367, opened 2016-02-22
Systemd to restart chef-client if it exited abnormally

mateuszkwiatkowski:


mateuszkwiatkowski:
Please disregard this PR as it's invalid. I'll follow up with correct version.

PR #372, opened 2016-03-14
Ensure upstart service picks up all config options

andrewjamesbrown:
Using upstart service does not reliably pick up the splay and interval
options.  This commit ensures that the options are picked up, and the
chef-client daemon is reliably restarted.  Simply using 'stop
chef-client ; start chef-client' results in a non-running daemon.
Combining the trap and nohup ensures the daemon runs with the new
config.

### DIFF
--- a/recipes/upstart_service.rb
+++ b/recipes/upstart_service.rb
@@ -37,6 +37,6 @@ service 'chef-client' do
   # This complication ensures (a) new splay/interval options are picked up and
   # (b) chef-client is running with the new config after restart is executed.
   #
-  restart_command 'nohup bash -c "trap SIGHUP SIGINT SIGTERM ; stop chef-client ; start chef-client"'
+  restart_command 'nohup bash -c "trap SIGHUP SIGINT SIGTERM ; stop chef-client ; sleep 1 ; start chef-client"'
   action [:enable, :start]
 end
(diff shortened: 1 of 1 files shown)


recipes/upstart_service.rb
```diff
@@ -31,5 +31,12 @@ class ::Chef::Recipe
 
 service 'chef-client' do
   provider Chef::Provider::Service::Upstart
+  # Upstart doesn't reliably restart chef-client with the new splay/interval options.
+  # Either the old config is maintained, or the chef-client daemon dies.
+  #
+  # This complication ensures (a) new splay/interval options are picked up and
+  # (b) chef-client is running with the new config after restart is executed.
+  #
+  restart_command 'nohup bash -c "trap SIGHUP SIGINT SIGTERM ; stop chef-client ; start chef-client"'
```
tonyflint: might want to add a 'sleep 1' between stop and start?

andrewjamesbrown: Fixed with 0e0ae1f

tas50:
We have some fixes merged into the core chef-client to handle correctly restarting upstart now since restart doesn't do what one would hope. I'm inclined to take the true fix there or to pass the stop followed by the start action. Doing it via bash with a sleep seems too fragile for this cookbook.

tas50:
I'm going to close this out since we fixed the real problem chef-client itself.

PR #380, opened 2016-04-01
Allow cmd arguments in cron definition

Igorshp:
Ran into an issue where I need to run chef-client on cron but with specific command line arguments.
Unfortunately there doesn't seem to be a way of doing this currently.

Here's the end result i'm after:

`/usr/bin/chef-client --audit-mode audit_only`

This PR allows us to add custom params to the end of the chef-client.

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -56,6 +56,7 @@ default['chef_client']['cron'] = {
   'path' => nil,
   'environment_variables' => nil,
   'log_file' => '/dev/null',
+  'cmd_args' => '',
   'append_log' => false,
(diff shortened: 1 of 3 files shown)


tas50:
@Igorshp If you're still interested in this can you rebase this change against the current state of the cookbook and remove the version bump since we have tooling. Also you'll want to document this change in the readme so other users can find out about it.

dheerajd-msys:
@Igorshp Please check ^

Igorshp:
Thanks, let me take a look

tas50:
Closing this out since we actually allow arbitrary arguments now.

Issue #387, opened 2016-04-22
client.rb.erb is problematic for external libraries supporting config options


mcquin:
The client.rb.erb template includes a reference to `Chef::Config`: https://github.com/chef-cookbooks/chef-client/blob/master/templates/default/client.rb.erb#L71-L73. This is problematic for external libraries which don't require `chef` but support loading configuration files like client.rb.

See: https://github.com/chef/ohai/issues/809

A workaround is to change to `ChefConfig::Config.from_file`, which is ultimately what `Chef::Config.from_file` is using.

PR #388, opened 2016-04-25
Use ChefConfig::Config.from_file to load config files

mcquin:
Resolves https://github.com/chef-cookbooks/chef-client/issues/387

\cc @tas50

### DIFF
--- a/templates/default/client.rb.erb
+++ b/templates/default/client.rb.erb
@@ -69,5 +69,9 @@ rescue NameError => e
 end
 
 Dir.glob(File.join(<%= node['chef_client']['conf_dir'].inspect %>, "client.d", "*.rb")).each do |conf|
+  <% if Gem::Version.new(Chef::VERSION) < Gem::Version.new('12.4.0') -%>
   Chef::Config.from_file(conf)
+  <% else -%>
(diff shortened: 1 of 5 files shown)


tas50:
@mcquin Will this work with Chef 11 releases?

mcquin:
Oh, probably not. Let me see what I can do about that.

tas50:
+1

mcquin:
moar review plz @thommay @iennae :)

templates/default/client.rb.erb
```diff
@@ -69,5 +69,9 @@ rescue NameError => e
 end
 
 Dir.glob(File.join(<%= node['chef_client']['conf_dir'].inspect %>, "client.d", "*.rb")).each do |conf|
-  Chef::Config.from_file(conf)
+  <% if Gem::Version.new(Chef::VERSION) < Gem::Version.new('12.4.0') -%>
+    Chef::Config.from_file(conf)
```
tas50: Can we avoid the indents here even though it makes the ERB harder to read

mcquin: You don't like the 4-space indent? :P

PR #390, opened 2016-05-04
Allow to use systemd timer instead of chef-client daemon mode

### LINKED ISSUE #358: Support for running as systemd timer
Hi,

It would be great if this cookbook supported running chef client as a systemd timer.
In addition, systemd > 229 supports "RandomizedDelaySec" which should allow for easy splay implementation.

What say you?

kamaradclimber:
### Description

This patches adds an attribute to configure a systemd timer starting the
chef-client.service.
chef-client.service is modified in this case not to run in daemon mode.

Options `interval` and `splay` are respected.
### Issues Resolved

Fix #358 
### Check List
- [x] All tests pass. See https://github.com/chef-cookbooks/community_cookbook_documentation/blob/master/TESTING.MD
- [x] The CLA has been signed. See https://github.com/chef-cookbooks/community_cookbook_documentation/blob/master/CONTRIBUTING.MD


kamaradclimber:
If you want I can go a bit further and make the systemd recipe uses the systemd_service provider from systemd cookbook.

kamaradclimber:
@tas50 any thoughts on this PR?

kamaradclimber:
@tas50, @thommay  would you have any feedback regarding this PR?

kamaradclimber:
test are currently failing bk of centos5 . I'll launch a rebuild to make them pass as they should

kamaradclimber:
@tas50 could you give some feedback on this MR (before I rebase it and fix the conflicts)?

kamaradclimber:
anyone? Sorry to ping you repeatedly @tas50  and @thommay . Could you give some feedback?
Thanks a lot

artem-sidorenko:
I miss this feature too, its a more elegant alternative to the cron way if you do not want to run chef-client in the daemon mode (e.g. because of transparent upgrades).

@kamaradclimber currently its a mixed implementation: one part via template, another part via systemd cookbook. In my eyes it makes sense to stick to one specific way: either to systemd cookbook or to manage the timer via template way if dependency to systemd cookbook is not welcome.

@tas50 it would be nice to get your view on this PR

kamaradclimber:
@artem-sidorenko I'd be happy to rework this PR if a maintainer express some interest in it.

kamaradclimber:
just in case, I've rebased the PR.

kamaradclimber:
I'll close the PR since it got no attention at all from maintainers.
:cry:

tas50:
Sorry for not replying here earlier. We did actually discuss the PR and the general concept is quite sound. We'd like to be able to do it without pulling in the massive systemd cookbook though. We have native systemd unit support now in chef and if we can get compat_resource to properly backport that I think that offers the best solution for everyone. We have the issue in this repo requesting the functionality and it's definitely something we're thinking about.

kamaradclimber:
Thanks @tas50  for your answer and sorry for not getting back sooner.

I just had a quick look at the systemd_unit resource in chef which seems really minimal (since we have to provide the content of the whole content of the unit).
Using systemd cookbook seems much more powerful.

I understand the need of being compatible but I really don't like the idea of compat_resource (if some users want recent chef features, updating chef-client is much more simpler than using a cookbook that monkeypatch many core chef classes).

Since I don't think I can convince you of accepting the PR as is, I'll close it.

artem-sidorenko:
@tas50 I have the similar view as @kamaradclimber on the `systemd_unit` resource of chef, sorry but this is not how it should be (example from [official docs](https://docs.chef.io/resource_systemd_unit.html)):

```ruby
systemd_unit 'sysstat-collect.timer' do
  enabled true
  content '[Unit]\nDescription=Run system activity accounting tool every 10 minutes\n\n[Timer]\nOnCalendar=*:00/10\n\n[Install]\nWantedBy=sysstat.service'
end
```

If I have to use this strange content think, I would either use the files/templates instead and would have the code more readable at this point.

I can fully understand the reasons of not accepting the dependency to the systemd cookbook.

What about the option of using the files/templates and rollout the systemd files via this way as already suggested? 

If you would accept a such PR, I would pick up the code of @kamaradclimber and add the change for files/templates. What do you think?

P.S. This is a second contribution from @kamaradclimber to the things maintained by Chef people ([here](https://github.com/test-kitchen/test-kitchen/pull/875) is another one), with hm, a bit suboptimal collaboration :-( Please do not get me wrong, I do not want to start any negative discussion around this topic, but just to highlight this fact for you.

artem-sidorenko:
btw, if you need help in the maintenance of this cookbook, I'm happy to support you

kamaradclimber:
@artem-sidorenko @tas50 I've thought (again!) about this PR.

Even if I don't like the reasonning behind compat_resource, I'm forced to recognize that it is your right as maintainer to target compatiblity with chef 12.1.
We do use a compat_resource fork internally that does nothing to chef core classes (to avoid nasty issues such as https://github.com/chef-cookbooks/compat_resource/issues/66) so adding compat_resource is not an issue for us.

Regarding systemd_unit, I hope it will get better in the future chef versions.

Here is an updated version of the PR using systemd_unit, I hope it can be useful for other chef users.

recipes/systemd_service.rb
```diff
@@ -2,6 +2,8 @@ class ::Chef::Recipe
   include ::Opscode::ChefClient::Helpers
 end
 
+include_recipe 'systemd::default'
```
artem-sidorenko: @kamaradclimber isn't this a leftover?

kamaradclimber: oups

kamaradclimber:
Anyway I could get feedback on this new implementation?

kamaradclimber:
@tas50 it seems that you don't have time to review incoming PR. Is there some process I should follow (or place I should go) to make progress?

kamaradclimber:
@tas50 I've implemented what you've asked, would you have some feedback?

Even something like "I don't have time to review now, but I'll try to have a look in the next month" would be better than silence.

kamaradclimber:
still no feedback?

iennae:
One other thing to note, with Chef 13 coming out in a few weeks, it might be useful to not include compat_resource and just make this a Chef 12.11+ with the next release. There is a lot of folks still using 12.5-7 but they can pin the version.

README.md
```diff
@@ -58,6 +59,7 @@ The following attributes affect the behavior of the chef-client program when run
 - `node['chef_client']['cron']['mailto']` - If set, `MAILTO` env variable is set for cron definition
 - `node['chef_client']['reload_config']` - If true, reload Chef config of current Chef run when `client.rb` template changes (defaults to true)
 - `node['chef_client']['daemon_options']` - An array of additional options to pass to the chef-client service, empty by default, and must be an array if specified.
+- `node['chef_client']['systemd']['timer']` - If true, uses systemd timer to run chef frequently instead of chef-client daemon mode (defaults to false)
```
kamaradclimber: good point. I'll fix that.

.kitchen.yml
```diff
@@ -78,6 +78,16 @@ suites:
...
   excludes: ["windows-2008r2", "windows-2012r2"]
 
+- name: timer_systemd
+  run_list:
+  - recipe[chef-client::systemd_service]
+  attributes:
+    chef_client:
+      systemd:
+        timer: true
+  includes: ["centos-7.2", "debian-8.2", "fedora-22", "fedora-23", "ubuntu-15.04"]
```
kamaradclimber: done, thanks

recipes/systemd_service.rb
```diff
@@ -15,20 +15,57 @@ class ::Chef::Recipe
...
+service_actions = [:enable]
+service_actions << :start unless node['chef_client']['systemd']['timer']
+
 service 'chef-client' do
   supports status: true, restart: true
-  action [:enable, :start]
+  action service_actions
+end
+
+systemd_unit 'chef-client.timer' do
```
kamaradclimber: good point, I'll fix that. In general it seems that most cookbooks are not "cleaning" resource when they are not necessary but in this case it would clearly make sense.

recipes/systemd_service.rb
```diff
@@ -15,20 +15,57 @@ class ::Chef::Recipe
...
+  content <<-EOH.gsub('^  ', '')
+  [Unit]
+  Description=chef-client periodic run
+
+  [Install]
+  WantedBy=timers.target
+
+  [Timer]
+  OnBootSec=1min
```
kamaradclimber: no particular reason, would you like to have it as an attribute? (maybe using the delay attribute?)

kamaradclimber:
tests are failing as in master. I'll try to rebase in a few days when tests will be passing again in master

recipes/systemd_service.rb
iennae: I'm divided about it being (yet another) attribute. I'm unfamiliar with this, so wondering if it's just a standard setting or something that folks want to configure.

iennae:
@kamaradclimber if you check Issue #358 you'll see that @tas50 updated the need for updating cookbook dependencies. 
Additionally, the standard policy is to support -6 months, so not including compact_resource is fine at this point to support older chef client versions.

recipes/systemd_service.rb
kamaradclimber: I think it can be made configurable later if someone ask for it. At the moment I don't know any users with a specific value

kamaradclimber:
Thanks a lot !

PR #396, opened 2016-05-21
Automatically restart systemd service on failure

sds:
### Description

The systemd service file provides a `Restart` option allowing you to
specify the conditions under which a service is automatically restarted.

Passing a value of `on-failure` ensures that the Chef client is
restarted in the event of a hard failure such as a segmentation fault or
OOM process killer.


sds:
The Travis failures look entirely unrelated:

```
An unexpected error occurred retrieving 'runit' (1.7.8) from the cookbook site
```

If there's something else that needs to be done (or more justification that needs to be provided) for this change to be merged, please let me know!

tas50:
Thanks @sds

Issue #397, opened 2016-05-25
chef-client recreates chef-client::task on each run


artisticcheese:
### Cookbook version
4.5.0

### Chef-client version

Chef: 12.10.24
### Platform Details
Windows 2012 R2

### Scenario:
chef-client::task recipe recreates task on each run if "frequency_modifier" is specified as default or overridden for environment

### Steps to Reproduce:
Runlist contains include_recipe 'chef-client::task'
Environment default attribute contains

```
{
  "chef_client": {
    "task": {
      "frequency_modifier": 17
    }
  }
}
```
### Expected Result:
Task shall not be recreated on each run

### Actual Result:
Each time recipe is run it recreates task 

```
Recipe: chef-client::task
  * directory[C:/chef/run] action create (up to date)
  * directory[C:/chef/cache] action create (up to date)
  * directory[C:/chef/backup] action create (up to date)
  * directory[C:/chef/log] action create (up to date)
  * directory[C:/chef] action create (up to date)
  * windows_task[chef-client] action create
Folder: \
HostName:                             CLDPOCCS0062
TaskName:                             \chef-client
Next Run Time:                        5/25/2016 8:59:00 AM
Status:                               Ready
Logon Mode:                           Interactive/Background
Last Run Time:                        5/25/2016 8:41:00 AM
Last Result:                          0
Author:                               SYSTEM
Task To Run:                          cmd /c " C:/opscode/chef/bin/chef-client   -L C:/chef/log/client.log   -c C:/chef/
client.rb -s 300 "
Start In:                             N/A
Comment:                              N/A
Scheduled Task State:                 Enabled
Idle Time:                            Disabled
Power Management:                     Stop On Battery Mode, No Start On Batteries
Run As User:                          SYSTEM
Delete Task If Not Rescheduled:       Disabled
Stop Task If Runs X Hours and X Mins: 72:00:00
Schedule:                             Scheduling data is not available in this format.
Schedule Type:                        One Time Only, Minute
Start Time:                           8:59:00 AM
Start Date:                           5/25/2016
End Date:                             N/A
Days:                                 N/A
Months:                               N/A
Repeat: Every:                        0 Hour(s), 17 Minute(s)
Repeat: Until: Time:                  None
Repeat: Until: Duration:              Disabled
Repeat: Stop If Still Running:        Disabled


```

tas50:
This should be fixed in 4.5.3

PR #398, opened 2016-05-27
Let the windows task escape the quotes and lose the stream redirection

mwrock:
This depends on https://github.com/chef-cookbooks/windows/pull/365 and will need to update the windows cookbook dependency constraint once that has been merged.

Currently the task recipe is not idempotent. See also this [discource discussion](https://discourse.chef.io/t/is-chef-client-task-recipe-supposed-to-be-idempotent/8540/7). With the changes to the windows cookbook, this should no longer need to escape the quotes of the task command since the task resource will manage that. Also this drops the stream redirection since the task (for reasons I dont know why) strips that from the task action.

### DIFF
--- a/metadata.rb
+++ b/metadata.rb
@@ -36,7 +36,7 @@ suggests 'runit'
 
 depends 'cron', '>= 1.7.0'
 depends 'logrotate', '>= 1.9.0'
-depends 'windows', '>= 1.39.0'
+depends 'windows', '>= 1.42.0'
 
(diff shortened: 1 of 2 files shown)


mwrock:
fixes #397

smurawski:
https://github.com/chef-cookbooks/windows/pull/365 has been merged and shipped in v1.42.0 of the windows cookbook.

mwrock:
Cool @smurawski I just bumped the windows dependency. I converged the `task-windows-2012r2` suite twice and validated that the second converge converges 0 resources and the verify passes.

PR #401, opened 2016-05-27
Use SYSTEM as the directory owner

### LINKED ISSUE #333: Windows root_owner should be System
Hello,

In windows the root equivalent user is `SYSTEM`, often we use the `Builtin\Administrator` account as a good alternative, because `SYSTEM` has at least the same right as `Administrator`
But ... I said **often**, and when you promote your server as a Domain Controller you loose the "Local Accounts" distinction, and `SYSTEM` starts to behave differently than `Administrator`!

Meaning that, if you promote your server as a Domain Controller, and chef is running as `SYSTEM`, currently you have big trouble to touch the `client.rb` file (via the config recipe for instance) because it's own by `Administrator`.

There is an helper to determine the [root_owner](https://github.com/opscode-cookbooks/chef-client/blob/master/libraries/helpers.rb#L37) but it's use an hardcoded SID + limit the query to `LocalAccount` which aren't there anymore :)

I know that new attributes are not accepted on this cookbook so ... I'll not create a PR adding a new `node['chef_client']['root_user']` attribute allowing me to use `SYSTEM` and not `Administrator` as **root_owner**.
Therefore, do you have any other solution?

Regards!

tas50:
### Description

We're currently trying to find a local admin account, but that won't exist on a DC.  SYSTEM will always exist and administrators group still has write on the dirs so this is really no change.

### Issues Resolved

#333 

### Check List
- [ ] All tests pass. See https://github.com/chef-cookbooks/community_cookbook_documentation/blob/master/TESTING.MD
- [ ] The CLA has been signed. See https://github.com/chef-cookbooks/community_cookbook_documentation/blob/master/CONTRIBUTING.MD

We're currently trying to find a local admin account, but that won't exist on a DC.  SYSTEM will always exist and administrators group still has write on the dirs so this is really no change.

### DIFF
--- a/libraries/helpers.rb
+++ b/libraries/helpers.rb
@@ -34,8 +34,8 @@ module Opscode
       end
 
       def root_owner
-        if ['windows'].include?(node['platform'])
-          wmi_property_from_query(:name, "select * from Win32_UserAccount where sid like 'S-1-5-21-%-500' and LocalAccount=True")
+        if platform?('windows')
+          'SYSTEM'
         else
           'root'
         end


mwrock:
What about using `administrator` and falling back to `system` if that user is not there? Just in case folks already using this expect `administrator`.

tas50:
Any tips on falling back in the Windows world?  How would I detect if that user exists?

smurawski:
While I like the idea of preserving compat - I like the PR as it stands.  If there is fallback, you'll have a different configuration on DCs than other servers in the environment.  And while the days of domains in the data center are numbered - it's still going to be a while before they all die.

mwrock:
yeah you should be able to do:
```
u=Chef::Util::Windows::NetUser.new("administrator")
u.get_info
```
if the user does not exist then the call to `get_info` will raise `Chef::Exceptions::UserIDNotFound`

smurawski:
👍

mwrock:
that sounds reasonable. Just wanted to bring it up and see if alarms sounded with anyone. :+1:

smurawski:
@mwrock one thing to be aware of - the difference in the wmi query to your netuser thing is that the wmi call will find re-named administrator accounts, where the netuser approach is reliant on the admin account not being renamed.

PR #406, opened 2016-06-08
Explicitly load files in client.d only for Chef 12.7 or older

sds:
Change the rendered `client.rb` file to include files in `client.d` _only if_ running a version of Chef < 12.8.

Chef 12.8+ includes support for automatically loading files in the `/etc/chef/client.d` directory [1]. This results in files being loaded twice if you are using the `chef-client` cookbook with a recent version of the Chef client, which can be problematic in certain situations (e.g. defining a custom formatter which is loaded twice results in duplicated output).

[1] https://docs.chef.io/release/12-8/release_notes.html

### DIFF
--- a/templates/default/client.rb.erb
+++ b/templates/default/client.rb.erb
@@ -68,6 +68,7 @@ rescue NameError => e
   Chef::Log.error e
 end
 
+<% if Gem::Version.new(Chef::VERSION) < Gem::Version.new('12.8') -%>
 Dir.glob(File.join(<%= node['chef_client']['conf_dir'].inspect %>, "client.d", "*.rb")).each do |conf|
   <% if Gem::Version.new(Chef::VERSION) < Gem::Version.new('12.4.0') -%>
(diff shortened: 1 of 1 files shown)


templates/default/client.rb.erb
```diff
@@ -68,10 +68,12 @@ rescue NameError => e
   Chef::Log.error e
 end
 
+<% if Gem::Version.new(Chef::VERSION) < Gem::Version.new('12.8') -%>
```
sds: I used `Gem::Version` because of a comment here: https://github.com/chef/chef/blob/05158511a3e85670f6937151a35705ff4a333053/lib/chef/version.rb#L30-L33

However, I might have misinterpreted. It sounds like in order to work with pre-release versions one should always use `Gem::Version`, which I've done here. If it makes more sense to use `Chef::Version`, I'm happy to change accordingly.

tas50:
Thanks @sds

PR #411, opened 2016-07-14
Remove support for daemontools and bluepill

tas50:
It's time to simplify this cookbook by supporting primarily OS default init systems.  Neither bluepill or daemontools gained widespread support and with systemd their need has mostly gone away.  We should remove the code necessary to support these init systems and further simplify this cookbook.


thommay:
:+1: LGTM

PR #419, opened 2016-08-15
Completes frequency_modifier explanation

spuder:


tas50:
Thanks for noticing that @spuder

PR #420, opened 2016-08-30
Don't explode under chef-solo.

coderanger:
Specifically `Chef::Config[:client_key]` is generally going to be nil under solo, and `exist?` doesn't like being passed nil. Regardless, this should be a no-op under solo.

### DIFF
--- a/recipes/delete_validation.rb
+++ b/recipes/delete_validation.rb
@@ -23,7 +23,13 @@ end
 
 # Don't do anything run running as solo. This can happen when using ChefSpec or
 # Test Kitchen.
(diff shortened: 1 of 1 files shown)


tas50:
This should probably log and return

coderanger:
What would it log? Most of the cases this would come up are testing related where any output would just be noise. I guess we could log only if not under TK or ChefSpec?

tas50:
We shouldn't silently skip things. Just throw a warn that it's skipping since solo won't have a file to delete

tas50:
Works for me

PR #421, opened 2016-09-01
Change path for rc.d script in FreeBSD

sekrett:
### Description

In FreeBSD /etc/rc.d is for system services only. Every init script installed by packages always goes to /usr/local. This is not a bug, but it violates the FreeBSD file system hierarchy. Other files you put it /etc are all correct. I did two commits, not sure if second is required.

### Check List
- [+] All tests pass. See https://github.com/chef-cookbooks/community_cookbook_documentation/blob/master/TESTING.MD
- [-] New functionality includes testing.
- [-] New functionality has been documented in the README if applicable
- [+] The CLA has been signed. See https://github.com/chef-cookbooks/community_cookbook_documentation/blob/master/CONTRIBUTING.MD

### DIFF
--- a/recipes/bsd_service.rb
+++ b/recipes/bsd_service.rb
@@ -29,6 +29,11 @@ when 'freebsd'
     mode 00755
   end
 
+  # Remove wrong rc.d script created by an older version of cookbook
+  file '/etc/rc.d/chef-client' do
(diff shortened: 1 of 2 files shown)


tas50:
Thanks @sekrett

PR #434, opened 2016-11-19
Doc: fixing information about cron attributes

artem-sidorenko:
### Description

There is no init_style cron. Lets have a link to the cron recipe documentation.


tas50:
Thanks

Issue #436, opened 2016-11-20
Everyone can't read chef.log


holmesb:
### Cookbook version
[7.0.0]

### Chef-client version
[12.16.42]

### Platform Details
[Ubuntu 14.04]

### Scenario:
Can't read chef.log as non-root\owner

### Steps to Reproduce:
Output logs to /var/log/chef/chef.log.  Try to open without sudo permission, eg less chef.log

### Expected Result:
Should be able to read log.

### Actual Result:
Can't.    Get error "_chef.log: Permission denied._"

attributes/default.rb has:
_# If log file is used, default permissions so everyone can read
default['chef_client']['log_perm'] = 00640_ 

If the intention is as per the comment (everyone to have read-access), needs to be:
default['chef_client']['log_perm'] = 00644_

tas50:
In general users consider these logs to be sensitive and we've chosen to default the config to 640 due to that. The permissions are and attribute if you wise to make them 644.

PR #437, opened 2016-11-20
[GH-436] - Making chef client logs readable by all

holmesb:
Fixes issue #436


tas50:
In general users consider these logs to be sensitive and we've chosen to default the config to 640 due to that. The permissions are and attribute if you wise to make them 644.

PR #443, opened 2016-12-02
Document / test setting a custom ohai plugin path

tas50:
Add a nil value attribute with a comment
Document the attribute in the readme
Simplify the logic in the template
Add a test for the content in the config suite

### DIFF
--- a/templates/default/client.rb.erb
+++ b/templates/default/client.rb.erb
@@ -40,7 +40,7 @@ https_proxy "<%= node["chef_client"]["config"]["https_proxy"] %>"
 no_proxy "<%= node["chef_client"]["config"]["no_proxy"] %>"
 <% end -%>
 
-<% if node.attribute?("ohai") && node["ohai"].attribute?("plugin_path") -%>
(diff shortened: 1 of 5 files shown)


templates/default/client.rb.erb
```diff
@@ -40,7 +40,7 @@ https_proxy "<%= node["chef_client"]["config"]["https_proxy"] %>"
 no_proxy "<%= node["chef_client"]["config"]["no_proxy"] %>"
 <% end -%>
 
-<% if node.attribute?("ohai") && node["ohai"].attribute?("plugin_path") -%>
+<% unless node["ohai"]["plugin_path"].nil? -%>
```
burtlo: You could write `if node["ohai"]["plugin_path"]` as `nil` and `false` are the only *false* values in Ruby. I do see the above more often in Chef recipes and my suggestion more in Ruby code.

tas50: Yep that would totally work. I was mostly just keeping things consistent with the other checks in the same template.

PR #453, opened 2016-12-29
Allow customizing the chef-client service name for Debian, SUSE, RHEL, and Fedora platform families.

jschripsema:
### Description

Expands the use of the 'svc_name' attribute, previously only used in the 'src_service' recipe, to Debian, SUSE, RHEL, and Fedora.

### Issues Resolved

Feature expansion. Fix to cookstyle quoting error in runit_service recipe.

### Check List

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -100,21 +100,25 @@ when 'aix'
   default['chef_client']['log_dir']     = '/var/adm/chef'
 when 'debian'
   default['chef_client']['init_style']  = node['init_package']
+  default['chef_client']['svc_name']    = 'chef-client'
   default['chef_client']['run_path']    = '/var/run/chef'
   default['chef_client']['cache_path']  = '/var/cache/chef'
   default['chef_client']['backup_path'] = '/var/lib/chef'
 when 'suse'
   default['chef_client']['init_style']  = 'systemd'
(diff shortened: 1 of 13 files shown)


attributes/default.rb
```diff
@@ -174,9 +178,9 @@
 default['chef_client']['log_rotation']['prerotate'] = nil
 default['chef_client']['log_rotation']['postrotate'] =  case node['chef_client']['init_style']
                                                         when 'systemd'
-                                                          'systemctl reload chef-client.service >/dev/null || :'
+                                                          "systemctl reload #{node['chef_client']['svc_name']}.service >/dev/null || :"
```
tas50: Since these are derived attributes this won't actually work. You'll end up with the default value of chef-client being used even if you override. You basically end up with the old behavior unfortunately.

tas50:
I'm actually going to close this out at this point. We're trying to walk back a lot of the complexity of this cookbook and this seems out the scope of the 95% use case we try to hit within community cookbooks. For almost every user there is just no reason to modify the chef-client service name. If that's truly necessary they can fork the cookbook and change the value.

https://github.com/chef-cookbooks/community_cookbook_documentation/blob/master/COOKBOOK_TEAM.MD#support-the-95-use-case

PR #455, opened 2017-01-16
Task Creds should be in an encrypted data bag

smcavallo:
Allow optional retrieval of config settings from encrypted data bag, specifically for scheduled task user and password (which should not be in clear text in node attributes.)

Signed-off-by: smcavallo <steve.cavallo@sonian.net>

### Description

Allow optional retrieval of config settings from encrypted data bag, specifically for scheduled task user and password (which should not be in clear text in node attributes.)

### Issues Resolved

[List any existing issues this PR resolves]

### Check List

- [X] All tests pass. See <https://github.com/chef-cookbooks/community_cookbook_documentation/blob/master/TESTING.MD>
- [X] New functionality includes testing.
   The windows boxes in the kitchen tests are marked private.  I do have tests which include adding an encrypted data bag to the test suites, if required.  The new functionality did not break any existing tests.
* I have tested this locally with custom windows images
- [X] New functionality has been documented in the README if applicable
- [X] All commits have been signed for the Developer Certificate of Origin. See <https://github.com/chef-cookbooks/community_cookbook_documentation/blob/master/CONTRIBUTING.MD>

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,6 +64,19 @@ The following attributes affect the behavior of the chef-client program when run
 - `node['chef_client']['task']['user']` - The user the scheduled task will run as, defaults to `'SYSTEM'`.
 - `node['chef_client']['task']['password']` - The password for the user the scheduled task will run as, defaults to `nil` because the default user, `'SYSTEM'`, does not need a password.
 
+Optionally, for better security, task user and password can be defined in a data bag
+- `node['chef_client']['data_bag']['name']` - = The data bag name, defaults to `'chef_client'`.
+- `node['chef_client']['data_bag']['config_item']` - The daat bag item, defaults to `'config'`.
+```json
+{
+  "task": {
+    "user": "scheduled_task_username",
+    "password": "scheduled_task_password"
+  },
+  "id": "config"
+}
+```
+
 The following attributes are set on a per-platform basis, see the `attributes/default.rb` file for default values.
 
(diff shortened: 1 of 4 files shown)


tas50:
@smcavallo I appreciate that you took the time to put this together, but in general we've avoided creating attributes to load credentials from encrypted data bags. We certainly want a more secure mechanism to pass credentials in, but that's not always an encrypted data bag. We end up with more and more attributes without actually achieving the flexibility of storage that is desired. 

I reworked the scheduled task configuration in this cookbook to instead use a new custom resource that setups up the credentials. The task recipe now calls that custom resource under the hood, which preserves backwards compatility, but also allows you to write your own wrapper recipe that calls the resource directory. How you conjure up the credentials to pass into that resource is up to you: encrypted databag, vault, chef-vault, carrier pidgin, etc. Let me know if I'm missing a use case here, but I think this solves your problem and moves us to a more resource driven cookbook at the same time.

https://github.com/chef-cookbooks/chef-client/pull/463

smcavallo:
@tas50 - thank you for the write up, and creating/merging a fix.  What you are saying makes sense, and you definitely understand what we are trying to achieve and have provided a better solution.  The provider method is a much better approach.  The only drawback is that it requires a wrapper cookbook, and we are trying to limit the number of wrapper cookbooks we have to maintain/support.  Despite that, your solution is a much cleaner and more flexible solution.

PR #458, opened 2017-01-26
skip validation.pem deletion in test-kitchen + chef-zero

Happycoil:
Signed-off-by: Tor Magnus Rakvåg <tm@intility.no>

### Description

This change prevents delete_validation from running when using chef-zero in test-kitchen. validation.pem is recreated on every converge in test-kitchen.

### Check List

### DIFF
--- a/recipes/delete_validation.rb
+++ b/recipes/delete_validation.rb
@@ -21,9 +21,9 @@ class ::Chef::Recipe
   include ::Opscode::ChefClient::Helpers
 end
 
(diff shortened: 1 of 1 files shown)


tas50:
I appreciate what you're trying to accomplish here, but we want to make sure that we can test things like the delete functionality under test kitchen. Having test kitchen take different code paths vs. a normal chef run is just opening things up to testing success with prod failure. That's something we'd really like to avoid so I'm going to close this one out. Thanks for taking the time to put it together though. It may be helpful to someone searching the repo.

PR #462, opened 2017-02-23
node_name is better than fqdn

lamont-granquist:
fqdn can be nil and the nil behavior this replaces is horrible

node_name has to be unique or nodes will be scribbling over the
same entries in the database.  also node_name has to be not-nil or
else chef-client itself throws an exception very early, so we
can now ignore that case.


lamont-granquist:
wasn't this red + busted?

Issue #460, opened 2017-01-27
windows task is created at each chef-client


lbn-chef-team:
Hello,

cookbook chef-client v7.1.0 / windows 2012r2 with chef client 12.17.44 

The windows task is created at each chef-client when using chef-client::task.

<pre>
  * windows_task[chef-client] action create
Folder: \
HostName:                             LBN-NT-DEV1
TaskName:                             \chef-client
Next Run Time:                        1/27/2017 10:00:00 PM
Status:                               Ready
Logon Mode:                           Interactive/Background
Last Run Time:                        N/A
Last Result:                          1
Author:                               Administrator
Task To Run:                          cmd /c " c:\linkbynet\programs\opscode\chef\bin\chef-client -L C:/chef/log/client.
log -c c:\chef/client.rb -s 3600 > NUL 2>&1"
Start In:                             N/A
Comment:                              N/A
Scheduled Task State:                 Enabled
Idle Time:                            Disabled
Power Management:                     Stop On Battery Mode, No Start On Batteries
Run As User:                          SYSTEM
Delete Task If Not Rescheduled:       Disabled
Stop Task If Runs X Hours and X Mins: 72:00:00
Schedule:                             Scheduling data is not available in this format.
Schedule Type:                        One Time Only, Hourly
Start Time:                           10:00:00 PM
Start Date:                           1/27/2017
End Date:                             N/A
Days:                                 N/A
Months:                               N/A
Repeat: Every:                        2 Hour(s), 0 Minute(s)
Repeat: Until: Time:                  None
Repeat: Until: Duration:              Disabled
Repeat: Stop If Still Running:        Disabled
SUCCESS: The scheduled task "chef-client" has successfully been created.
</pre>

I've did some test and it seems that it is the escape characters ^ used at:
https://github.com/chef-cookbooks/chef-client/blob/master/recipes/task.rb
line 43.

when removing this <pre>^> NUL 2^>^&1</pre> the task is not re-created at each chef-client anymore.

Regards,

jugatsu:
Confirmed.
This is because windows_task parses command from schtasks.exe /XML output for idempotence.
https://github.com/chef-cookbooks/windows/blob/master/providers/task.rb#L188
But xml output of ^> transforms to `&gt;` and so & to `&amp;`

jugatsu:
https://github.com/chef-cookbooks/windows/issues/426

jugatsu:
[...]

PR #466, opened 2017-03-21
Make chef_client_scheduled_task idempotent

jugatsu:
### Description
Remove command redirections from chef-client command as `windows_task`
doesn't support them, as per https://github.com/chef-cookbooks/windows/issues/426

### Issues Resolved

https://github.com/chef-cookbooks/chef-client/issues/460

### DIFF
--- a/resources/scheduled_task.rb
+++ b/resources/scheduled_task.rb
@@ -45,7 +45,7 @@ action :add do
   start_time = new_resource.frequency == 'minute' ? (Time.now + 60 * new_resource.frequency_modifier).strftime('%H:%M') : nil
   windows_task 'chef-client' do
     run_level :highest
-    command "cmd /c \"#{client_cmd} ^> NUL 2^>^&1\""
+    command "cmd /c \"#{client_cmd}\""
 
     user               new_resource.user
     password           new_resource.password


jugatsu:
@tas50 are there any chances to be merged?

rosario-mattera:
Hi,
I've deleted the redirection from the line 43 but the task is still created each time.  How is it possible? On which system have you tested this feature?
I'm using a win2012 r2.

Regards,
Rosario

jugatsu:
@rosario-mattera which version of cookbook do you use?

jugatsu:
can you paste output to gist?

rosario-mattera:
@jugatsu the version is 7.1.0. I've moved the recipe's update on the production environment and I see that, after the bootstrap of the node, which essentially executes all the recipes included in the run-list, in order to have an idempotent execution of the task recipe, just one further execution is needed and then it becomes idempotent. So I think that this unique execution is needed to manage the transition from the "redirected" version of the task to the "not-redirected". 

```
Parameters	Initial State	Final State
task_name	\chef-client	chef-client
command	cmd /c " C:/opscode/chef/bin/chef-client -L C:/chef/log/client.log -c C:/chef/client.rb -s 300 > NUL 2>&1"	cmd /c " C:/opscode/chef/bin/chef-client -L C:/chef/log/client.log -c C:/chef/client.rb -s 300"
user	SYSTEM	SYSTEM
run_level		highest
frequency_modifier		120
frequency		minute
start_time		21:42
```

jugatsu:
@rosario-mattera so it works as expected on v8.0.1 of cookbook?

rosario-mattera:
I expect that the behavior on v8.0.1 is an idempotent recipe. The same happens with this fix on v7.1.0 but if I have an already deployed chef-client, it requires one more further execution to be idempotent.

Thanks again,
Rosario

PR #467, opened 2017-03-28
Fix nil attribute in cron recipe

ncerny:
### Description

node['node_name'] is not a valid ohai attribute, and does not pass unit tests.  This change switches to node['fqdn'] to avoid a nil attribute.

### Check List

### DIFF
--- a/recipes/cron.rb
+++ b/recipes/cron.rb
@@ -87,7 +87,7 @@ end
 
 # Generate a uniformly distributed unique number to sleep.
 if node['chef_client']['splay'].to_i > 0
-  checksum   = Digest::MD5.hexdigest(node['node_name'])
+  checksum   = Digest::MD5.hexdigest(node['fqdn'])
   sleep_time = checksum.to_s.hex % node['chef_client']['splay'].to_i
 else
   sleep_time = nil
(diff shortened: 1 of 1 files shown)


recipes/cron.rb
```diff
@@ -87,7 +87,7 @@ class ::Chef::Recipe
 
 # Generate a uniformly distributed unique number to sleep.
 if node['chef_client']['splay'].to_i > 0
-  checksum   = Digest::MD5.hexdigest(node['node_name'])
+  checksum   = Digest::MD5.hexdigest(node['fqdn'])
```
coderanger: Okay, so using the FQDN here is dangerous because during a DNS failure it can spontaneously change. In newer Chef we have `node['shard_seed']` coming from Ohai as a stable, uniform seed value for this kind of thing but since we need fallback for older versions we should probably check that first and then use a hash of `node.name` as that's the easiest stable source for most users. If they want something more configurable, they can upgrade Chef and get the full power of the shard_seed plugin :)

```ruby
seed = node['shard_seed'] || Digest::MD5.hexdigest(node.name).to_s.hex
sleep_time = seed % node['chef_client']['splay'].to_i
```

iennae:
Hey @ncerny check out @coderanger's comments about using fqdn. Thanks for investigating this error and working towards making builds green again!

recipes/cron.rb
ncerny: I like it.

PR #472, opened 2017-04-05
use shell_out and remove Chef 10 code

lamont-granquist:
SANE_PATHS was removed in 0.10.10

foodcritic is entirely correct in yelling about not using shell_out


tas50:
I was eyeing removing this path junk earlier. Thanks for the PR

PR #477, opened 2017-04-20
Properly start systemd timer when necessary

kamaradclimber:
Otherwise it requires a reboot of the node to activate the timer.

Also don't enable the service when using timer.

### Check List

### DIFF
--- a/recipes/systemd_service.rb
+++ b/recipes/systemd_service.rb
@@ -68,8 +68,8 @@ systemd_unit 'chef-client.timer' do
   EOH
   enabled timer
   if timer
-    action [:create, :enable]
+    action [:create, :enable, :start]
   else
-    action [:disable, :delete]
+    action [:stop, :disable, :delete]
   end
 end


kamaradclimber:
I'll submit a second PR tomorrow to improve migration from service -> time (and possible timer -> service)

kamaradclimber:
all tests pass except linting that does not seem related to my patch.

tas50:
@kamaradclimber Any chance you can drop the migration work into this PR. It would be great to get both of these issues fixed and then cut a release

artem-sidorenko:
> I'll submit a second PR tomorrow to improve migration from service -> time (and possible timer -> service

@kamaradclimber can you please have a look at GH-484? Maybe it addresses this point (at least partly)

kamaradclimber:
@artem-sidorenko that possible. Frankly I hadn't take time to work on this topic in the last weeks.

GH-484 might solve the issue but we should make sure that stopping the service when timer is enabled does not break the current chef-client run.

tas50:
This was merged via #492

PR #483, opened 2017-04-29
Systemd timeout

artem-sidorenko:
### Description
It can be used in the systemd timer setup to avoid stalled chef runs.
E.g. chef run gets stalled sometimes if used on the notebook and it gets
moved to another network during the run. The stalled chef process does not allow
the execution of any other chef runs.
Using the systemd timeout we can kill this stalled chef runs via systemd.

### Issues Resolved

Can be seen as solution for https://github.com/chef-cookbooks/chef-client/issues/320, but for systemd timer setup instead of cron.

### Check List

### DIFF
--- a/recipes/systemd_service.rb
+++ b/recipes/systemd_service.rb
@@ -42,6 +42,11 @@ service_unit_content = {
 
 service_unit_content['Service'].delete('Restart') if timer
 
+if node['chef_client']['systemd']['timeout']
+  service_unit_content['Service']['TimeoutSec'] =
+    node['chef_client']['systemd']['timeout']
+end
+
 systemd_unit 'chef-client.service' do
   content service_unit_content
   action :create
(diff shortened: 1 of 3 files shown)


tas50:
@artem-sidorenko Can you rebase this for the changes to how we configure the systemd unit now?

artem-sidorenko:
@tas50 done, I also added a small fix on top when timer gets disabled

recipes/systemd_service.rb
```diff
@@ -42,6 +42,11 @@ class ::Chef::Recipe
 
 service_unit_content['Service'].delete('Restart') if timer
 
+if node['chef_client']['systemd']['timeout']
```
iennae: based on documentation, it seems like this needs to check to make sure that timer isn't a oneshot and if it is do a WARN as this setting would be ignored in that case.

artem-sidorenko: @iennae

From the [official docs](https://www.freedesktop.org/software/systemd/man/systemd.service.html#TimeoutStartSec=):

> Defaults to DefaultTimeoutStartSec= from the manager configuration file, except when Type=oneshot is used, in which case the timeout is disabled by default

Maybe I get it wrong, from my understanding it means that oneshot services does not have timeout per default, but everything else does. So setting it this way should be ok?

artem-sidorenko:
@iennae @tas50 anything missing here?

PR #503, opened 2017-07-07
Add support for ClearLinux

albertomurillo:
### Description
* Set systemd as init type when the platform is clearlinux
* Create /etc/systemd/system directory (clearlinux is a stateless
  os and in fresh deploys the directory might not exists.)

For more information about clearlinux visit clearlinux.org

### Check List


recipes/systemd_service.rb
```diff
@@ -13,7 +13,8 @@ class ::Chef::Recipe
   ['fedora'] => ['fedora', 'sysconfig', 'chef-client'],
   ['rhel'] => ['redhat', 'sysconfig', 'chef-client'],
   ['suse'] => ['redhat', 'sysconfig', 'chef-client'],
-  ['debian'] => ['debian', 'default', 'chef-client']
+  ['debian'] => ['debian', 'default', 'chef-client'],
+  ['clearlinux'] => ['clearlinux', 'chef', 'chef-client']
```
iennae: Is 'chef' correct here? I don't think we have a clearlinux image to test against, but this would be for the service configuration.

albertomurillo: Yes, clearlinux has a different approach to configuraiton files and they told me to put chef configuration files under `/etc/chef` rather than `/etc/sysconfig/`. The code above generates `/etc/chef/client.rb` file which makes more sense to them than `/etc/sysconfig/client.rb`

iennae: Can you provide me current version information so I can write up tests? Thanks!

PR #510, opened 2017-07-30
Do not refer to the deprecated cookbook

akitada:
### Description

Refer to chef_client_updater instead of omnibus_updater (deprecated)


tas50:
Yep. Thanks

Issue #490, opened 2017-05-17
Issue with Chef Client 13  and the `chef_client_scheduled_task` resource on Windows


seththoenen:
### Cookbook version
chef-client >= 8.1.0

### Chef-client version
chef-client 13.0.118

### Platform Details
Windows Server 2008 R2
Windows Server 2012 R2

### Scenario:
The `chef_client_scheduled_task` resource converges on every run on chef-client cookbook version >= 8.1.0 when paired with chef-client 13.0.118 on Windows Server 2008 R2 and Windows Server 2012 R2. Currently utilizing the windows cookbook version 3.0.5 (the latest at the time of this writing). Please refer to the table below:

![image](https://cloud.githubusercontent.com/assets/524558/26165388/67ab24c8-3af5-11e7-92d6-4dcde0af6c77.png)

### Steps to Reproduce:
Run the chef-client cookbook version >= 8.1.0 on Windows Server 2008 R2 or Windows Server 2012 R2 paired with the windows cookbook = 3.0.5 and chef-client 13.0.118

### Expected Result:
The `chef_client_scheduled_task` resource should not converge on every chef-client run.

### Actual Result:
The `chef_client_scheduled_task` resource converges on every chef-client run.

iennae:
8.0.2 had known issues with working on Windows systems. We'll dig into the issue with 8.1.0 and Chef 13. Thanks for reporting this issue.

iennae:
This could be due to a idempotence issue with the windows_task resource https://github.com/chef/chef/issues/6093. 

Could you provide more information about how you are specifying the resource itself? Are you specifying the start time?

seththoenen:
I'm using the default attributes that come with the chef-client cookbook. I'm not overriding any functionality or calling any resources directly at this time. I'm simply calling the cookbook:

`include_recipe 'chef-client::default'`

So, to answer your question, I'm not specifying  a start time.

Thanks!

username-is-already-taken2:
I noticed this frozen error today with 2012 R2 on an Azure VM installed with the chef extension.  I'll try upgrading the cookbook

seththoenen:
I'm still having issues with the latest version of the Chef-client cookbook (version 8.1.6). The Windows scheduled task will still converge. I'm using the default settings that come with the cookbook.

Chef-Client version: 13.1.31

Windows Server Versions:
2008 R2 Standard
2008 R2 Enterprise
2012 Standard
2012 R2 Standard
2016 Standard
[...]

PR #522, opened 2017-09-14
[MSYS-661] [In-Progress]Fix chef_client_scheduled_task idempotence

dheerajd-msys:
### Description
Updated cookbook for making `chef_client_scheduled_task` resource idempotent.

### Issues Resolved
Fixes https://github.com/chef-cookbooks/chef-client/issues/490

### DIFF
--- a/recipes/task.rb
+++ b/recipes/task.rb
@@ -45,4 +45,5 @@ chef_client_scheduled_task 'Chef Client' do
   log_directory node['chef_client']['log_dir']
   chef_binary_path node['chef_client']['bin']
   daemon_options node['chef_client']['daemon_options']
+  not_if { ::File.exist?("C:\\Windows\\System32\\Tasks\\#{node['chef_client']['task']['task_name']}") }
 end
--- a/recipes/task.rb
+++ b/recipes/task.rb
@@ -45,5 +45,4 @@ chef_client_scheduled_task 'Chef Client' do
   log_directory node['chef_client']['log_dir']
   chef_binary_path node['chef_client']['bin']
   daemon_options node['chef_client']['daemon_options']
-  not_if { ::File.exist?("C:\\Windows\\System32\\Tasks\\#{node['chef_client']['task']['task_name']}") }
 end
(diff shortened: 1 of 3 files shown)


seththoenen:
Would the scheduled task resource update if someone wanted to change the functionality of the scheduled task? For example, would a new task be created if one wanted to change the interval or splay?

dheerajd-msys:
Per current `chef-client` cookbook `chef_client_scheduled_task` resource will not update the interval or splay.
I have made it idempotent for only if `scheduled task` is created already.

dheerajd-msys:
@btm @tas50 , Please review.

tas50:
Won't this prevent the task from ever updating if the options or other properties change?

recipes/task.rb
```diff
@@ -45,4 +45,5 @@ class ::Chef::Recipe
   log_directory node['chef_client']['log_dir']
   chef_binary_path node['chef_client']['bin']
   daemon_options node['chef_client']['daemon_options']
+  not_if { ::File.exist?("C:\\Windows\\System32\\Tasks\\#{node['chef_client']['task']['task_name']}") }
```
btm: As @tas50 indicated, this would prevent the task from being updated.

dheerajd-msys: Yes @btm, I've mentioned it too in my above comment. To make the resource `chef_client_scheduled_task` idempotent I was needed to add this check.

dheerajd-msys: @tas50 @btm, This cookbook calculates the `start_time` for frequency `:minute` which seems inappropriate as per `windows_task` resource description. `start_time` needs to be mentioned only for `:once` 
```
frequency

    Ruby Type: Symbol | Default Value: :hourly

    Frequency with which to run the task. Values include: :minute, :hourly, :daily, :weekly, :monthly, :once, :on_logon, :onstart, :on_idle. The :once value requires the start_time property.
``` 
Please check here https://github.com/chef/chef-web-docs/blob/master/chef_master/source/resource_windows_task.rst#properties

I am trying to modify the recipe like gist shared here https://gist.github.com/dheerajd-msys/e771890bc8bf0f515582c77bba84c7ae.

Please suggest your views for modifications.

NimishaS:
@tas50 , can you please confirm the reason of adding `start_time` in case of `minutes` frequency: https://github.com/chef-cookbooks/chef-client/blob/master/resources/scheduled_task.rb#L45? Does removing it has any bad impact?

We checked the commit history of the file, but the reason isn't very clear from the commit messages.

dheerajd-msys:
We have identified a bug in windows_task resource for `frequency :monthly` that it is not idempotent. Also found that for `frequency :once` , `frequency_modifier` in `windows_task` option should not be provided. We will be fixing it soon on another PR so closing this now.

PR #530, opened 2017-11-09
partial fix chef_client_schedule_task

dheerajd-msys:
### Description

This PR  is a partial fix for making `chef_client_schedule_task` resource idempotent. Testing is remaining.  

### Issues Resolved
Partial fixes https://github.com/chef-cookbooks/chef-client/issues/490

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -68,9 +68,11 @@ default['chef_client']['systemd']['restart'] = 'always'
 
 # Configuration for Windows scheduled task
 default['chef_client']['task']['frequency'] = 'minute'
-default['chef_client']['task']['frequency_modifier'] = node['chef_client']['interval'].to_i / 60
+default['chef_client']['task']['frequency_modifier'] = 30
 default['chef_client']['task']['user'] = 'SYSTEM'
-default['chef_client']['task']['password'] = nil # Password is only required for none system users
+default['chef_client']['task']['password'] = '' # Password is only required for none system users
+default['chef_client']['task']['task_name'] = 'chef-client'
+default['chef_client']['task']['start_time'] = '06:30'
 
 default['chef_client']['load_gems'] = {}
 
--- a/resources/scheduled_task.rb
+++ b/resources/scheduled_task.rb
@@ -19,12 +19,12 @@
 
 resource_name :chef_client_scheduled_task
 
-property :user, String, default: 'System'
+property :user, String, default: node['chef_client']['task']['user']
 property :password, String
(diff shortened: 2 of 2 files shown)


NimishaS:
@tas50 , @btm, while trying to make `frequency: monthly` idempotent, we came across some issues.
If `frequency_modifier: 4` is set with frequency `:monthly`, the schedule task sets `months: Apr, Aug, Dec` in the XML. 
So in order to set the `current_resource.frequency_modifier`, we can use some formula like `12 / months_arr.length`. 
But there's also an attribute called `months` which also sets the same XML `months` element.
Suppose the user writes a recipe:
```
windows_task "my_task" do
  frequency   :monthly
  months       "Jan, Mar"
end
``` 
This will also set `current_resource.frequency_modifier` if we apply the logic that I mentioned above. 
We cannot be sure whether the existing scheduled task was created by passing `frequency_modifier` attribute or `months` attribute in case of `frequency :monthly`. 
It's difficult to make the resource idempotent in this case.
The solution to this problem can be to allow setting only one attribute for `frequency :monthly`... either `frequency_modifier` or `months`.

resources/scheduled_task.rb
```diff
@@ -19,12 +19,12 @@
 
 resource_name :chef_client_scheduled_task
 
-property :user, String, default: 'System'
+property :user, String, default: node['chef_client']['task']['user']
```
tas50: These node attributes won't actually work if someone overrides them in a role/env/cookbook.

attributes/default.rb
```diff
@@ -68,9 +68,11 @@
 
 # Configuration for Windows scheduled task
 default['chef_client']['task']['frequency'] = 'minute'
-default['chef_client']['task']['frequency_modifier'] = node['chef_client']['interval'].to_i / 60
+default['chef_client']['task']['frequency_modifier'] = 30
 default['chef_client']['task']['user'] = 'SYSTEM'
-default['chef_client']['task']['password'] = nil # Password is only required for none system users
+default['chef_client']['task']['password'] = '' # Password is only required for none system users
+default['chef_client']['task']['task_name'] = 'chef-client'
+default['chef_client']['task']['start_time'] = '06:30'
```
NimishaS: Is there any reason to update these attributes for fixing https://github.com/chef-cookbooks/chef-client/issues/490?

tas50:
I'm closing this out since this bandaids the fix in a way that I'm uncomfortable doing and also switches from resource driven to cookbook driven config logic, which we're trying to avoid.

PR #534, opened 2017-11-29
Update systemd_service.rb

tocomo:
added daemon_options to systemd config as well becase the chef_client daemon_options are currently not injected into systemd config file
Signed-off-by: Tobias Strauss <tac@gmx.li>


tas50:
Merging under the obvious fix policy. Thanks for this contribution @tocomo

PR #561, opened 2018-03-09
handle mixed lines if log_location is nil

glensc:
### Description

Fixes `log_location ssl_verify_mode :verify_none` (yes on same line) when `node['chef_client']['config']['log_location'] = nil`

The first fix is to remove "omit newline" in `.erb`: `%>` vs `-%>`, that would result:
```
log_location
ssl_verify_mode :verify_none
```

which is not syntax error anymore, but i've updated the if-condition so that the `nil` gets rendered as `nil` in output, which is more correct aesthetically.

```
log_location nil
ssl_verify_mode :verify_none
```

### Check List


tas50:
Thanks

PR #562, opened 2018-03-09
handle empty (but not nil) handlers block

glensc:
### Description

Currently when all handlers are empty, the try-catch block is still written because the condition is made on `.nil?` check, but [default is defined as empty array: `[]`](https://github.com/chef-cookbooks/chef-client/blob/v10.0.0/attributes/default.rb#L79-L81).

### Check List


tas50:
Nice catch. Thanks

Issue #563, opened 2018-03-17
systemd ignores ['chef_client']['ca_cert_path'] in /etc/sysconfig/chef-client when set on rhel


bcg62:
`systemd: Ignoring invalid environment assignment 'export SSL_CERT_FILE=/etc/pki/tls/certs/ca-bundle.crt': /etc/sysconfig/chef-client`

I suspect the export needs to be removed

https://github.com/chef-cookbooks/chef-client/blob/master/templates/default/redhat/sysconfig/chef-client.erb#L16

bcg62:
Doing it the same as suse:

```
<% if node["chef_client"]["ca_cert_path"] %>
SSL_CERT_FILE="<%= node["chef_client"]["ca_cert_path"] %>"
export SSL_CERT_FILE
<% end %>
```
https://github.com/chef-cookbooks/chef-client/blob/master/templates/default/suse/sysconfig/chef-client.erb

seems to correct the issue with no systemd warnings.

tested on centos 7.4.1708

tas50:
Fixed in 10.0.2

PR #564, opened 2018-03-17
systemd ignoring ca_cert_path on rhel

bcg62:
addressing #563


tas50:
Thanks @bcg62. I Released this as 10.0.2

PR #566, opened 2018-03-28
Fix the handling of log_locations that aren't strings

pwalz:
Signed-off-by: Peter Walz <pnw@umn.edu>

### Description

In Chef 12.9, `:syslog` and `:win_evt` were added as supported log locations in `client.rb`, adding to the existing `STDOUT`, `STDERR`, and file-based logging. Presently, the only locations supported by `chef-client::config` are `STDOUT` and file-based. (Chef interprets the strings `syslog` and `win_evt` as their corresponding symbols, but we should be able to use the symbols directly.)

This PR fixes `templates/default/client.rb.erb` so all of the `log_location` values are rendered correctly.

### Issues Resolved

See above.

### Check List


tas50:
Thanks @pwalz

PR #569, opened 2018-04-13
Fix failures on Windows with Chef 14

tas50:
We have a fix in master that restores the resources method, but we should unbreak Chef 14.0.190 users. This does exactly that and actually cleans up the code a bit too

### DIFF
--- a/libraries/helpers.rb
+++ b/libraries/helpers.rb
@@ -46,8 +46,7 @@ module Opscode
         d_owner = root_owner
         %w(run_path cache_path backup_path log_dir conf_dir).each do |dir|
           # Do not redefine the resource if it exist
-          next if find_resource(:directory, node['chef_client'][dir])
-          directory node['chef_client'][dir] do
+          find_resource(:directory, node['chef_client'][dir]) do
             recursive true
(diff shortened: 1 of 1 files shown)


libraries/helpers.rb
```diff
@@ -46,15 +46,12 @@ def create_chef_directories
...
-            resources(directory: node['chef_client'][dir])
-          rescue Chef::Exceptions::ResourceNotFound
-            directory node['chef_client'][dir] do
-              recursive true
-              mode '0755' if dir == 'log_dir'
-              owner d_owner
-              group node['root_group']
-            end
+          next if find_resource(:directory, node['chef_client'][dir])
+          directory node['chef_client'][dir] do
```
lamont-granquist: find_resource with a block actually does exactly what you want:

```
find_resource(:directory, node['chef_client'][dir]) do
            recursive true
            mode '0755' if dir == 'log_dir'
            owner d_owner
            group node['root_group']
end
```

no need to guard it.

https://github.com/chef/chef/blob/master/lib/chef/dsl/declare_resource.rb#L187-L190

tas50: Updated. Thanks

PR #573, opened 2018-05-10
Handlers should support single instance of class

smcavallo:
### Description

This fixes a nasty bug where handlers are called multiple times during chef runs.
For example when the following is defined -
```
exception_handlers << SimpleReport::UpdatedResources.new
```
This adds a new instance of this class to the Chef::Config -
https://github.com/chef/chef/blob/92d85cf4c67680d778cbe8bc5c676f9a10fa1442/lib/chef/handler.rb#L110-L113
If you inspect what is being stored in this value, it is a new instance of this class upon every converge.
This causes each instance of the handler to fire when the handler is called.
So, for example, a notification handler would fire once on the first chef converge, twice on the second converge, three times on the third converge, etc
This is certainly not the expected behavior.  Additionally it's probably not healthy to the internals of chef to be storing this many instances of a class - mem leak.

This PR attempts to fix that by creating a new instance of the class, and only adding the instantiated instance to the handlers.  The handlers will only ever be added to the Chef::Config and be executed once.

I didn't want to break backwards compatibility so the new (undocumented) node attribute was added.
My preference would be to replace the current functionality and write out the client.rb in the new way for all users.  I can't imagine a scenario where we would want to instantiate a new instance of the class for each chef run and the memory/performance impact that would come with that.  That may be considered a breaking change for this cookbook.
As such I did not update the README.

### Check List

- [X ] All tests pass. See <https://github.com/chef-cookbooks/community_cookbook_documentation/blob/master/TESTING.MD>
- [X ] New functionality includes testing.
- [X ] All commits have been signed for the Developer Certificate of Origin. See <https://github.com/chef-cookbooks/community_cookbook_documentation/blob/master/CONTRIBUTING.MD>

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -80,6 +80,7 @@ default['chef_client']['load_gems'] = {}
 default['chef_client']['config']['start_handlers'] = []
 default['chef_client']['config']['report_handlers'] = []
 default['chef_client']['config']['exception_handlers'] = []
+default['chef_client']['config']['handlers'] = []
 
 # If set to false, changes in the `client.rb` template won't trigger a reload
 # of those configs in the current Chef run.
--- a/recipes/config.rb
+++ b/recipes/config.rb
@@ -84,6 +84,7 @@ template "#{node['chef_client']['conf_dir']}/client.rb" do
     chef_config: node['chef_client']['config'],
     chef_requires: chef_requires,
     ohai_disabled_plugins: node['ohai']['disabled_plugins'],
+    handlers: node['chef_client']['config']['handlers'],
     start_handlers: node['chef_client']['config']['start_handlers'],
     report_handlers: node['chef_client']['config']['report_handlers'],
     exception_handlers: node['chef_client']['config']['exception_handlers']
--- a/resources/scheduled_task.rb
+++ b/resources/scheduled_task.rb
@@ -19,8 +19,8 @@
 
 resource_name :chef_client_scheduled_task
 
-property :user, String, default: 'System', sensitive: true
-property :password, String, sensitive: true
+property :user, String, default: 'System'
+property :password, String
 property :frequency, String, default: 'minute', equal_to: %w(minute hourly daily monthly once on_logon onstart on_idle)
 property :frequency_modifier, [Integer, String], default: 30
 property :start_date, String, regex: [%r{^[0-1][0-9]\/[0-3][0-9]\/\d{4}$}]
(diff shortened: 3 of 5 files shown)


tas50:
Ping @lamont-granquist here

lamont-granquist:
so if the daemonized chef-client process is re-reading its Chef::Config every time through the interval loop and then the forked child processes are inheriting the Chef::Config that might be a cause of this (about the only one that I can think of) but that is a bug in core chef, not really a bug in this cookbook.

smcavallo:
@lamont-granquist -  I think you're right that this is not addressing the root cause of the issue.
This is an issue on all linux OS we're using.

You can see the behavior when adding additional logging ->
https://github.com/chef/chef/blob/63b7d9e4fb250afb697d7d151cdcc102d9b258ac/lib/chef/handler.rb#L112
https://github.com/chef/chef/blob/63b7d9e4fb250afb697d7d151cdcc102d9b258ac/lib/chef/handler.rb#L121

During each converge it creates a new instance of the class (in my case Journalist) and appends that new instance to the existing array of handlers in Chef::Config[:report_handlers]
You can see the this in the logs.
That falls in line with what you said - "the daemonized chef-client process is re-reading its Chef::Config every time"

You can see in the output - during every run it gets a new instance of the class and each instance of the class handles the event.
in client.rb -
`report_handlers << Journalist.new(%w[template])`
```
[2018-10-24T16:49:36+00:00] INFO: Chef Run complete in 22.336394182 seconds
[2018-10-24T16:49:36+00:00] INFO: Running report handlers
[2018-10-24T16:49:36+00:00] INFO: HANDLER: #<Journalist:0x0000000004aa47c8>
[2018-10-24T16:49:36+00:00] INFO: Writing the updated resource report ...
[2018-10-24T16:49:36+00:00] INFO: HANDLER: #<Journalist:0x0000000004ab54d8>
[2018-10-24T16:49:36+00:00] INFO: Writing the updated resource report ...
[2018-10-24T16:49:36+00:00] INFO: Report handlers complete
```

Another way it could be possibly handled in chef client instead of clearing the config, for any handler which is a class instance, it could dispose of that specific instance.  that way, you'd always get a new instance of the class and that instance would only live as long as the chef run.

I do agree it should be handled via chef client.  I tried to fix it here instead because if during a chef run the chef client is going to always instantiate a new instance of the class and that's the normal behavior then it's what's passed into chef should change.

smcavallo:
closing since issue is with chef-client

PR #586, opened 2018-09-15
Fix warning about ownership of client.log on Windows platform

jugatsu:
Signed-off-by: Anton Kvashenkin <anton.jugatsu@gmail.com>

### Description

Fix warning about ownership of `client.log` file on Windows platform.

```
* file[C:/chef/log/client.log] action create[2018-09-15T02:38:19-07:00] WARN: Mode 640 includes bits for the owner, but owner is not specified
       [2018-09-15T02:38:19-07:00] WARN: Mode 640 includes bits for the group, but group is not specified
        (up to date)
```

### DIFF
--- a/recipes/config.rb
+++ b/recipes/config.rb
@@ -49,8 +49,14 @@ end
 # libraries/helpers.rb method to DRY directory creation resources
 create_chef_directories
 
+# We need to set these local variables because the methods aren't
+# available in the Chef::Resource scope
+d_owner = root_owner
+
 if log_path != 'STDOUT'
   file log_path do
+    owner d_owner
+    group node['root_group']
     mode node['chef_client']['log_perm']
   end
 end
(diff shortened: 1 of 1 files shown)


attributes/default.rb
```diff
@@ -44,7 +44,7 @@
 default['chef_client']['log_dir']     = '/var/log/chef'
 
 # If log file is used, default permissions so everyone can read
-default['chef_client']['log_perm'] = '640'
+default['chef_client']['log_perm'] = '0644'
```
tas50: We probably don't want 644 for the chef-client log. That's a pretty significant security concern given what's in that log file.

jugatsu: You are correct. That's probably copy-paste error from `template` resource :)

jugatsu: With `640` `Users` can read `client.log` file.


 
Without `owner` and `group` `Users` can read `client.log` also.



@tas50 What do you think?

jugatsu: On linux platform


BTW I wil fix comment in this PR as everyone can't read `client.log` on Linux. Only `root` and `root` group.
https://github.com/chef-cookbooks/chef-client/blob/master/attributes/default.rb#L46

jugatsu:
@tas50 Done.

tas50:
Thanks @jugatsu

Issue #409, opened 2016-06-23
Add "on boot" option for applicable recipes


docwhat:
### Suggestion:

I'd like chef-client to run immediately on booting.

In cron, this can be done with `@reboot` and Windows tasks have a similar trigger.

Obviously, services wouldn't need or want this.

The use cases:
- Some recipes have to reboot after running.  This lets Chef pick up where it left off after the reboot (linux kernel changes, some windows things).
- If a system is down for a while it should run Chef ASAP to get compliant again.

thomasrutkowski:
This makes several of us. 
I've been searching for a way to do just this and am currently looking at having cron kick of a converge just after reboot. 
Are there any strong contraindications to force a converge via cron? Perhaps I haven't looked in the right places, but there's seems to be very little information on the subject.

docwhat:
Nope. Works great. 

I use it for all our Linux boxes.

willdearborn:
I thought I saw a notification about a new option to run a converge on daemon startup, but I'm not finding it.  I would also like this option, if it doesn't exist.

Stromweld:
For windows with the new recommended task instead of service the windows_task provider should be enhanced to allow :onstart with recurring options so the default chef-client task can be setup to run right away on boot as well as every 30 minutes etc... 

Currently I have chef-client setup with defaults and have a recipe creating a second task to accomplish the initial run on boot. Unfortunately the frequency when changed from :minute to :onstart doesn't set the recurring options in windows task scheduler.

windows_task 'chef-client_onstartup' do
  command 'cmd.exe /c c:/opscode/chef/embedded/bin/chef-client.bat'
  cwd 'C:'
  user 'SYSTEM'
  run_level :highest
  frequency :onstart
  action [:create, :enable]
end

dheerajd-msys:
Closing this issue based on https://github.com/chef-cookbooks/chef-client/pull/590#issuecomment-425682125 and https://github.com/chef-cookbooks/chef-client/pull/590#issuecomment-425708499.

PR #590, opened 2018-09-19
Add onstart property to chef_client_scheduled_task resource

jugatsu:
With this new property we can control creation of additional scheduled task to run chef-client on Windows boot.

Signed-off-by: Anton Kvashenkin <anton.jugatsu@gmail.com>

### Issues Resolved

Partially resolves https://github.com/chef-cookbooks/chef-client/issues/409

I think it would be better to use an additional property `onstart` rather than `frequency` with `onstart` value because in this case we can use only _one_ `chef_client_scheduled_task` resource.

### DIFF
--- a/resources/scheduled_task.rb
+++ b/resources/scheduled_task.rb
@@ -21,8 +21,9 @@ resource_name :chef_client_scheduled_task
 
 property :user, String, default: 'System', sensitive: true
 property :password, String, sensitive: true
-property :frequency, String, default: 'minute', equal_to: %w(minute hourly daily monthly once on_logon onstart on_idle)
+property :frequency, String, default: 'minute', equal_to: %w(minute hourly daily monthly once on_logon on_idle)
 property :frequency_modifier, [Integer, String], default: 30
+property :onstart, [true, false], default: false
 property :start_date, String, regex: [%r{^[0-1][0-9]\/[0-3][0-9]\/\d{4}$}]
 property :start_time, String, regex: [/^\d{2}:\d{2}$/]
 property :splay, [Integer, String], default: 300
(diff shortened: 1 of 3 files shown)


resources/scheduled_task.rb
```diff
@@ -60,9 +61,24 @@
     start_time         start_time_value
     start_day          new_resource.start_date unless new_resource.start_date.nil?
   end
+
+  if new_resource.onstart
+    windows_task "#{new_resource.task_name}-onstart" do
+      run_level :highest
```
tas50: This will need to be gated based on chef release since only the latest versions of chef support this.

jugatsu: I don't get it. Which property only supported by latest chef release? It was tested with `13.10` and it works like a charm.

jugatsu: @tas50 can you give feedback please

jugatsu:
@tas50 I think we can merge this PR as it works on any version of `chef-client`.

tas50:
I'm not really a fan of those change actually. The goal of shipping the resource was to allow us to use it in the recipe with minimal attributes in a way that supported 95% of the use cases. I think we've done that. For users that fall outside that use case they have the resource and they can use it directly. In this should just able to to use the resource twice and have 1 of the frequencies set to onstart. We've trying very hard to avoid attribute bloat in these recipes by supporting every possible action.

jugatsu:
I agree with you. I'm ok with my current wrapper cookbook.

```ruby
include_recipe 'chef-client::config'
include_recipe 'chef-client::default'
include_recipe 'chef-client::delete_validation' unless ENV['TEST_KITCHEN']

# configure starting chef-client on Windows boot
chef_client_scheduled_task 'Chef Client on start' do
  user node['chef_client']['task']['user']
  frequency 'onstart'
  frequency_modifier 1
  splay 60
  config_directory node['chef_client']['conf_dir']
  daemon_options node['chef_client']['daemon_options']
  task_name 'chef-client-on-start'
  notifies :reboot_now, 'reboot[Ensure node is rebooted after chef-client cookbook run]', :immediate
  only_if { platform?('windows') }
  action :add
end

reboot 'Ensure node is rebooted after chef-client cookbook run' do
  only_if { node['gf_chef_client']['reboot_after_onstart'] }
  action :nothing
end
```

PR #591, opened 2018-09-22
Express special strings for log_location as corresponding symbol

dwmarshall:
### Description

This change converts the configuration attribute `log_location` from either the string `"syslog"` or `":syslog"` to the symbol `:syslog`. If the actual `log_location` setting in `client.rb` is `"syslog"`, then it works just like `:syslog`. However, if the string is `":syslog"`, `chef-client` will do what is probably intended in daemon mode but not in the foreground.

In any event, when someone sets the attribute to the string `":syslog"`, it is probably the case that they want the symbol `:syslog`. Also, someone who sees the string `"syslog"` in `client.rb` might be confused as to what is really happening.

Also, I tilted at my anti-bundler sentiment a bit by changing the comments in `Gemfile`. Bundler and ChefDK do not play well together. I am willing to discard that commit if desired.

### Issues Resolved

N/A

### Check List


jugatsu:
hello @tas50. Can you review this PR please. We lack support of setting `:winevt` or `win_evt` as string in `chef-client` cookbook. As a result we can't set `default['chef_client']['config']['log_location'] = ':win_evt'` in Policifile.rb per policy group, for example. Because Policfile.rb doesn't support setting attributes as symbols.

tas50:
Thanks for the tests for this one @dwmarshall

PR #595, opened 2018-10-10
Fix duplicate daemon options in systemd unit

### LINKED ISSUE #594: Setting daemon_options causes the options to appear twice in systemd unit files
### Cookbook version

11.0.1

### Chef-client version

14.1.12

### Platform Details

RedHat 7

### Scenario:

When setting daemon_options the chef-client daemon is started with the parameters passed twice.
In https://github.com/chef-cookbooks/chef-client/blob/master/recipes/systemd_service.rb#L50
```
exec_options = if timer
                 '-c $CONFIG $OPTIONS'
               else
                 '-c $CONFIG -i $INTERVAL -s $SPLAY $OPTIONS'
end

service_unit_content = {
    ...
    'ExecStart' => "#{client_bin} #{exec_options} #{node['chef_client']['daemon_options'].join(' ')}",
    ...
}
```

It is clear that $OPTIONS from sysconfig and daemon_options are used. Since the environment file $OPTIONS are used for init.d and others too daemon_options should be removed here.

When daemon_options are setting the formatter this causes duplicate log output due to this bug: https://github.com/chef/chef/issues/7737

### Steps to Reproduce:

Set any daemon_options and the daemon is started with these options duplicated.

### Expected Result:

Options should appear only once.

### Actual Result:

See https://github.com/chef/chef/issues/7737

joerg:
This is a workaround to fix https://github.com/chef/chef/issues/7737 in chef-client.

Signed-off-by: Joerg Herzinger <joerg.herzinger@oiml.at>

### Description

Daemon options appear only once in systemd unit files.

### Issues Resolved

#594

### Check List


lamont-granquist:
is this is a real fix to remove duplicate options or is this dropping functionality here in this cookbook that other people need, because of a bug in chef?

joerg:
Sorry for this confusion, and no, this does not drop any functionality at all. The daemon_options will still be there, but just not duplicated.
In the systemd unit resource you can see that the enviornments file gets included (https://github.com/chef-cookbooks/chef-client/blob/master/recipes/systemd_service.rb#L49) and in the environments file the daemon options are set anyway (https://github.com/chef-cookbooks/chef-client/blob/master/templates/default/redhat/sysconfig/chef-client.erb#L14). So this just drops the duplicate daemon_options which just happened to cause a "bug", or more of a nuisance, for me.

lamont-granquist:
cool, that's a better explanation of what is going on here... :+1:

tas50:
Thanks @joerg

PR #599, opened 2018-11-27
Stop chef-client windows service after creating scheduled task

jasonwbarnett:
### Description

Along with disabling the service from stopping we should actually stop the windows chef-client service.

### Issues Resolved

N/A

### Check List

### DIFF
--- a/recipes/task.rb
+++ b/recipes/task.rb
@@ -35,12 +35,6 @@ client_bin = find_chef_client
 Chef::Log.info("Using chef-client binary at #{client_bin}")
 node.default['chef_client']['bin'] = client_bin
 
-windows_service 'chef-client' do
-  startup_type :disabled
-  action :configure_startup
-  only_if { ::Win32::Service.exists?('chef-client') }
-end
-
 chef_client_scheduled_task 'Chef Client' do
   user node['chef_client']['task']['user']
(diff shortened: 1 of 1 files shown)


jasonwbarnett:
@tas50 does this make sense? It seemed odd that there wasn't parity between the linux cron recipe and this one. Thoughts?

tas50:
I think you're going to run into some nasty issues here where the chef-client laying this down is the one trying to stop itself. This order *might* make that work, but any chance you can test that somehow?

jasonwbarnett:
Yeah, it throws a `Timeout::Error`, ugh. I wish we could somehow stop the service cleanly. Let me play with this a bit and I'll get back to you.

jasonwbarnett:
The only thing I can think to do to "trick" chef into stopping chef from the scheduled task is to do something like,

```ruby
windows_service 'chef-client' do
  action :stop

  only_if { ::File.exist?('c:/chef/.stop_chef') }
end

file 'c:/chef/.stop_chef'
```

jasonwbarnett:
You know... we could always just use the `ignore_failure true` on the service stop.

jasonwbarnett:
@tas50 
yikes... was this ready?

PR #604, opened 2018-12-20
Reload launchd on change

pludi:
### Description

`notify` the `chef-client` service on macOS, if running as a daemon, when the daemon-plist changes.

### Issues Resolved

none


hawksight:
Looks like it would fix the issue #620 I raised.

tas50:
Thanks @pludi

PR #612, opened 2019-01-25
Fix order of conditional tests in client.rb.erb template

### LINKED ISSUE #611: client.rb.erb template does not check for nil handlers properly
### Cookbook version
11.0.5

### Chef-client version
14.9.13 (inside test kitchen)

### Platform Details
`$ chef --version`
Chef Development Kit Version: 3.6.57
chef-client version: 14.8.12
delivery version: master (5fb4908da53579c9dcf4894d4acf94d2d9ee6475)
berks version: 7.0.7
kitchen version: 1.24.0
inspec version: 3.2.6

### Scenario:
Default config attribute in hash k,v format. Example:
```ruby
default['chef-client']['config'] = {
  rubygems_url: 'https://example.com'
}
```
Fails to converge.  When setting a config attribute like so:
```ruby
default['chef-client']['config']['rubygems_url'] = 'https://example.com'
```
Converges cleanly.

### Steps to Reproduce:
1. Generate cookbook `chef generate cookbook chef-client-config`
2. Add `depends 'chef-client'` to `metadata.rb`.
3. Configure a chef-client config attribute
```ruby
# attributes/default.rb
default['chef-client']['config'] = {
  rubygems_url: 'https://example.com'
}
```

4. Include the recipe
```ruby
# recipes/default.rb
include_recipe 'chef-client::config'
```

5. Setup a test kitchen
```yaml
# .kitchen.yml
---
driver:
  name: vagrant

provisioner:
  name: chef_zero
  always_update_cookbooks: true

verifier:
  name: inspec

platforms:
  - name: centos-7

suites:
  - name: default
    run_list:
      - recipe[chef-client-config::default]
    verifier:
      inspec_tests:
        - test/integration/default
    attributes:
```

5. Converge the test kitchen `kitchen converge`

### Expected Result:
Clean converge

### Actual Result:
```
           ================================================================================
           Error executing action `create` on resource 'template[/etc/chef/client.rb]'
           ================================================================================
           
           Chef::Mixin::Template::TemplateError
           ------------------------------------
           undefined method `empty?' for nil:NilClass
           
           Resource Declaration:
           ---------------------
           # In /tmp/kitchen/cache/cookbooks/chef-client/recipes/config.rb
           
            80: template "#{node['chef_client']['conf_dir']}/client.rb" do
            81:   source 'client.rb.erb'
            82:   owner d_owner
            83:   group node['root_group']
            84:   mode '0644'
            85:   variables(
            86:     chef_config: node['chef_client']['config'],
            87:     chef_requires: chef_requires,
            88:     ohai_disabled_plugins: node['ohai']['disabled_plugins'],
            89:     start_handlers: node['chef_client']['config']['start_handlers'],
            90:     report_handlers: node['chef_client']['config']['report_handlers'],
            91:     exception_handlers: node['chef_client']['config']['exception_handlers']
            92:   )
            93: 
            94:   if node['chef_client']['reload_config']
            95:     notifies :run, 'ruby_block[reload_client_config]', :immediately
            96:   end
            97: end
           
           Compiled Resource:
           ------------------
           # Declared in /tmp/kitchen/cache/cookbooks/chef-client/recipes/config.rb:80:in `from_file'
           
           template("/etc/chef/client.rb") do
             action [:create]
             default_guard_interpreter :default
             source "client.rb.erb"
             declared_type :template
             cookbook_name "chef-client"
             recipe_name "config"
             owner "root"
             group "root"
             mode "0644"
             variables {:chef_config=>{"rubygems_url"=>"https://example.com", "log_location"=>"/var/log/chef/client.log"}, :chef_requires=>[], :ohai_disabled_plugins=>[], :start_handlers=>nil, :report_handlers=>nil, :exception_handlers=>nil}
             path "/etc/chef/client.rb"
             verifications []
           end
           
           Template Context:
           -----------------
           on line #50
            48: <% end -%>
            49: 
            50: <% if !@start_handlers.empty? || !@report_handlers.empty? || !@exception_handlers.empty? -%>
            51: # Do not crash if a handler is missing / not installed yet
            52: begin
           
           System Info:
           ------------
           chef_version=14.9.13
           platform=centos
           platform_version=7.6.1810
           ruby=ruby 2.5.3p105 (2018-10-18 revision 65156) [x86_64-linux]
           program_name=/opt/chef/bin/chef-client
           executable=/opt/chef/bin/chef-client
           
       
       Running handlers:
       [2019-01-25T22:37:24+00:00] ERROR: Running exception handlers
       Running handlers complete
       [2019-01-25T22:37:24+00:00] ERROR: Exception handlers complete
       Chef Client failed. 0 resources updated in 02 seconds
       [2019-01-25T22:37:24+00:00] FATAL: Stacktrace dumped to /tmp/kitchen/cache/chef-stacktrace.out
       [2019-01-25T22:37:24+00:00] FATAL: Please provide the contents of the stacktrace.out file if you file a bug report
       [2019-01-25T22:37:24+00:00] FATAL: Chef::Mixin::Template::TemplateError: undefined method `empty?' for nil:NilClass
```

ngharo:
### Description
Checks handlers for nil then empty, in that order. Prevents a blow up when handlers are indeed nil as seen in #611.

### Issues Resolved
Fixes #611 

### Check List

### DIFF
--- a/templates/default/client.rb.erb
+++ b/templates/default/client.rb.erb
@@ -47,20 +47,20 @@ ohai.plugin_path << "<%= node["ohai"]["plugin_path"] %>"
 ohai.disabled_plugins = [<%= @ohai_disabled_plugins.map { |k| k.match(/:/) ? k.gsub(/^:/, '').to_sym.inspect : k.inspect }.join(",") %>]
 <% end -%>
 
-<% if !@start_handlers.empty? || !@report_handlers.empty? || !@exception_handlers.empty? -%>
+<% if !@start_handlers.nil? || !@report_handlers.nil? || !@exception_handlers.nil? -%>
 # Do not crash if a handler is missing / not installed yet
 begin
-<% unless @start_handlers.nil? -%>
+<% unless @start_handlers.empty? -%>
 <% @start_handlers.each do |handler| -%>
   start_handlers << <%= handler["class"] %>.new(<%= handler["arguments"].join(',') %>)
 <% end -%>
 <% end -%>
-<% unless @report_handlers.nil? -%>
+<% unless @report_handlers.empty? -%>
 <% @report_handlers.each do |handler| -%>
   report_handlers << <%= handler["class"] %>.new(<%= handler["arguments"].join(',') %>)
 <% end -%>
 <% end -%>
-<% unless @exception_handlers.nil? -%>
+<% unless @exception_handlers.empty? -%>
 <% @exception_handlers.each do |handler| -%>
   exception_handlers << <%= handler["class"] %>.new(<%= handler["arguments"].join(',') %>)
 <% end -%>


templates/default/client.rb.erb
```diff
@@ -47,20 +47,20 @@ ohai.plugin_path << "<%= node["ohai"]["plugin_path"] %>"
 ohai.disabled_plugins = [<%= @ohai_disabled_plugins.map { |k| k.match(/:/) ? k.gsub(/^:/, '').to_sym.inspect : k.inspect }.join(",") %>]
 <% end -%>
 
-<% if !@start_handlers.empty? || !@report_handlers.empty? || !@exception_handlers.empty? -%>
+<% if !@start_handlers.nil? || !@report_handlers.nil? || !@exception_handlers.nil? -%>
```
lamont-granquist: most likely all that should happen here is that the original `empty?` check here should be deleted along with its associated `end`.

the way this will work after the change is that 2 out of 3 of these could be nil, but as long as one isn't, then it'll enter through this if statement and then later those nil values will blow up in their `#each` statement.

it also isn't necessary at all to protect with `empty?` checks like this since `each` just does nothing when passed an empty array.

you could change the conditional so that if any of the arrays were `empty?` that it would not fire (`unless @start_handlers.nil? || @report_handlers.nil? || @exception_handlers.nil?`) but then if you have one nil value the other non-nil values do not get rendered which is another kind of bug.

so its best to do the `nil?` checks down below around the loops, and then let `#each` deal with the `empty?` case implicitly.

tas50:
This is fixed in #635

Issue #557, opened 2018-02-23
Test that systemd timer configuration is working


mattouille:
### Cookbook version
10.0.0

### Chef-client version
13.7.16

### Platform Details
Ubuntu 16.04

### Scenario:
Determine if timers are set correctly according to the SystemD spec

given chef-client.timer
then /etc/systemd/system/chef-client.timer exists (another systemd unit file) exists
then /etc/systemd/system/chef-client.service exists
then /etc/systmd/system/chef-client.timer Interval should correlate to attribute
then /etc/systmd/system/chef-client.timer SPlay should correlate to attribute
then chef-client.service IS installed
then chef-client.service IS NOT enabled
then chef-client.service IS NOT running

### Steps to Reproduce:
Set attributes 

```
default['chef-client']['init_style'] = 'systemd'
default['chef_client']['systemd']['timer'] = true
```

### Expected Result:
Tests run according to the above scenario

### Actual Result:
No tests are currently done

### Note:
The functionality is working, but it's working on a hope basis. No test is being run to determine the actual outcome, therefore, changes in SystemD could potentially break this cookbook.

dheerajd-msys:
This issue is fixed in https://github.com/chef-cookbooks/chef-client/pull/614.

@mattouille Please close this issue.

PR #614, opened 2019-01-30
fix systemd timer activation

### LINKED ISSUE #613: OnActiveSec is the wrong timer property for triggering the service unit on an interval
### Cookbook version
11.0.5

### Chef-client version
14.9.13

### Platform Details
CentOS 7.6.1810

### Scenario:
systemd timer unit triggered activation of chef-client.service

### Steps to Reproduce:
set up the timer/service units as the cookbook currently does 

e.g.
```
PRODUCTION [root@bastion.prod.iad01.treehouse chef-client]$ cat /etc/systemd/system/chef-client.timer
[Unit]
Description = chef-client periodic run

[Install]
WantedBy = timers.target

[Timer]
OnBootSec = 1min
OnActiveSec = 1800sec
RandomizedDelaySec = 900sec
PRODUCTION [root@bastion.prod.iad01.treehouse chef-client]$ cat /etc/systemd/system/chef-client.service 
[Unit]
Description = Chef Client daemon
After = network.target auditd.service

[Service]
Type = oneshot
EnvironmentFile = /etc/sysconfig/chef-client
ExecStart = /usr/bin/chef-client -c $CONFIG $OPTIONS
ExecReload = /bin/kill -HUP $MAINPID
SuccessExitStatus = 3

[Install]
WantedBy = multi-user.target
```

start the timer, wait for service unit to be activated at the $interval mark. service unit will not be activated once $interval has elapsed unless the timer unit is restarted during the chef converge.

you can see from `systemctl list-timers` output whether it will work in the future, an n/a for "LEFT" or "NEXT" means it will not be triggered.

### Expected Result:
service unit is activated every $interval, not just once after $interval from timer unit starting

### Actual Result:
service unit is only activated once after timer unit is activated

nathwill:
### Description

makes timer based activation work again

### Issues Resolved

#613 #557 

### Check List

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,6 +30,12 @@ env:
   - CHEF_VERSION=14 INSTANCE=service-systemd-fedora-latest
   - CHEF_VERSION=14 INSTANCE=service-systemd-opensuse-leap
   - CHEF_VERSION=14 INSTANCE=service-systemd-ubuntu-1604
+  - CHEF_VERSION=14 INSTANCE=timer-systemd-debian-8
+  - CHEF_VERSION=14 INSTANCE=timer-systemd-debian-9
+  - CHEF_VERSION=14 INSTANCE=timer-systemd-centos-7
+  - CHEF_VERSION=14 INSTANCE=timer-systemd-fedora-latest
+  - CHEF_VERSION=14 INSTANCE=timer-systemd-ubuntu-1604
+  - CHEF_VERSION=14 INSTANCE=timer-systemd-opensuse-leap
   - CHEF_VERSION=14 INSTANCE=service-init-centos-6
   - CHEF_VERSION=14 INSTANCE=service-init-amazonlinux
   - CHEF_VERSION=13 INSTANCE=cron-ubuntu-1604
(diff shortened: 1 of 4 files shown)


nathwill:
pushing this to our cluster as a hot-fix to get our nodes converging again, will report back.

nathwill:
verified this is working in our cluster; testing this in kitchen is a little more complicated... poking around at unit properties to see if there's a way to verify that the timer has triggered and has an upcoming trigger time

nathwill:
ok, so we can check NextElapseUSec properties for verifying that a sd-event has been scheduled and the timer isn't going to be idle indefinitely: https://github.com/systemd/systemd/blob/2d479ff1cc045eae971687ee770720fc444dee37/src/core/timer.c#L39

we can also check LastTriggerUSec properties to see if the timer has triggered, but there's a race between the OnBootSec and the integration tests running that is usually won by the integration tests, so without inserting a sleep longer than OnBootSec we can't really verify that NextElapseUSec has been calculated after the OnBootSec trigger has been applied (i.e. based on OnUnitInactiveSec). at the point where we're inserting sleeps into the tests, i'm inclined to say it's not worth it, but in this case, this has actually been an issue for some users, so... :man_shrugging: thoughts?

nathwill:
well, at least these tests are better than nothing. unfortunately it looks like it's pretty tough to reliably get information about trigger timing beyond the initial one, so leaving that out for now.

voroniys:
I'm having the issue as well. Please include this fix ASAP, since current implementation is incorrect!

tas50:
Sorry for the delay on this one @nathwill and thanks for the tests

PR #619, opened 2019-03-20
Invert the conditional to be more accurate.

jjustice6:
### Description

The conditional in the `cron` cookbook (which this still uses) uses `node['os']`, let's make the check match exactly.

### Issues Resolved

No open issue, but this causes it to fail on Solaris (among others).

### Check List

- [X] All commits have been signed for the Developer Certificate of Origin. See <https://github.com/chef-cookbooks/community_cookbook_documentation/blob/master/CONTRIBUTING.MD>

### DIFF
--- a/recipes/cron.rb
+++ b/recipes/cron.rb
@@ -121,8 +121,9 @@ if node['chef_client']['cron']['use_cron_d']
     command cmd
   end
 else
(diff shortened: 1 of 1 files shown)


tas50:
This is totally the right way to do this. Thanks

PR #625, opened 2019-04-19
Expose ability to accept chef license

tyler-ball:
Adding ability to accept upcoming Chef license

### Description

This is a backwards compatible way to prepare for the upcoming Chef
license change. Set the attribute `node['chef_client']['chef_license'] =
'accept'` (or `'accept-no-persist'` if you do not want persistence and
will always specify this attribute) to accept the Chef license. When you
upgrade to Chef Client 15 the license will be automatically accepted and
there will not be any interruption to service.

Requires running the 'chef_client::config' recipe to work correctly. If
you do not run this recipe add `chef_license "accept"` to your client
config in whatever way you manage that.

### Issues Resolved

N/A

### Check List

### DIFF
--- a/templates/default/client.rb.erb
+++ b/templates/default/client.rb.erb
@@ -12,7 +12,6 @@ end
 <% unless @chef_license.nil? -%>
 chef_license "<%= @chef_license %>"
 <% end -%>
-
 <% @chef_config.keys.sort.each do |option| -%>
   <% next if %w{ node_name exception_handlers report_handlers start_handlers http_proxy https_proxy no_proxy }.include?(option) -%>
   <% case option -%>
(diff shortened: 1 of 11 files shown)


tyler-ball:
Should I instead add the attribute under `node['chef_client']['config']['chef_license']` since it is only applied by the config recipe?

templates/default/client.rb.erb
```diff
@@ -8,6 +8,11 @@
 end
 
 <% end -%>
+
```
tas50: One other thing. This is going to result in extra spaces being added even if the config isn't templated out which causes updates of every chef run when someone bumps their cookbook. People tend to freak out about that so puttings the spacing inside the unless is helpful here.

PR #626, opened 2019-04-29
Fix systemd unit transition from service to timer and vice-versa

Annih:
### Description

This is not a common scenario, but I hit an issue when transitioning my Chef systemd config from a service to a timer: the chef-client service and the current process is terminated before setuping the timer.

When debugging the issue, I also tested the reverse scenario, transitioning from timer to service and I hit another bug: chef tries to restart the systemd_timer it has just removed.

This change fix above issues.

### Check List


tas50:
Thanks @Annih

PR #630, opened 2019-06-10
Fix incorrect attribute in README for chef_license

americanhanko:
### Description

Fixes the `README.md` to accurately reflect the attribute that needs to be set in order to accept the chef license.

### Issues Resolved

Since `node['chef_client']['license']` is not implemented, setting that attribute obviously did not write to the `client.rb` as expected.


tas50:
Thanks @americanhanko

PR #640, opened 2019-10-23
set the client_fork default to nil

lamont-granquist:
this shouldn't be set in the config.rb to either true or false.

the default sets it to --fork which forces --fork to cli use
of chef-client as well, which only causes additional issues for
CLI use where the supervisor process is not necessary or
required.

the use of the --interval flag or --once or similar flags correctly
gets the default setting of fork/no-fork right and that behavior
inside the client should be left alone.

a concrete example of the kind of problems this causes is that running the
chef_client_updater cookbook from the command line will send SIGKILL to the
PPID if it is running with client_fork.  that is the correct behavior, but
for some reason that can SIGKILL the shell running chef-client as well, which
is some kind of weird process-group leader issue -- one which i don't have the
time to research so the best solution is that CLI invocation of the client
should never fork (because the supervisor process on a CLI invocation is super
duper 100% useless).  but with `client_fork true` that means everyone needs to know
to use `--no-fork` on every command line invocation -- but forking works fine for
99% of the time until you hit the use cases where it doesn't.

See also: https://github.com/chef/chef/pull/6914

Since that was considered a possibly breaking change it went into 14.0 and has been in
all 14.x and 15.x behavior.  Since the prior behavior of 13.x was to have client_fork
turned on all the time by default, those users will see no behavioral change at all from
this PR.

Users can of course force this back to `true`, but I would highly discourage that.

If something breaks downstream, please submit PRs to add `--fork` or `--no-fork` correctly to the
default command arguments of how the ultimate supervisor process invokes chef-client (ie. systemd config or
cronjob settings for the CLI command rather than the global chef config.rb setting).


lamont-granquist:
hopefully some tests fail in useful ways...

kitchen.yml
```diff
@@ -21,6 +21,7 @@ platforms:
   - name: debian-10
   - name: fedora-29
```
tas50: You can just change this to fedora-latest and we'll never need to update it again

kitchen.yml
```diff
@@ -21,6 +21,7 @@ platforms:
...
   - name: freebsd-12
+  - name: opensuse-leap-15
   - name: sles-11-sp2
```
tas50: this can go away now that SLES 11 is EOL

lamont-granquist:
```
242319  kitchen verify service-systemd-opensuse-leap-15
242320  kitchen verify service-systemd-amazonlinux-2
```

those both succeed on my desktop so i can't replicate the travis failures.  i can only speculate something like the travis VMs are being too slow and verification is running before the service has come up or some kind of race condition like that.

PR #654, opened 2020-03-18
Add log_file_name property to chef_client_scheduled_task

tas50:
This way it doesn't use any attributes


phiggins:
Does this break stuff that rely on the old behavior?

tas50:
Yep this is going to be a breaking change in a major release

PR #663, opened 2020-03-18
Prevent failures when setting certain frequency types

tas50:
random_delay and frequency_modifier are not supported for all the possible frequency types. Our own tests in this cookbook were actually failing to run based on the frequency value we set. This should prevent those failures by not passing in unsupported properties. I extracted it to helper methods so I can test this easily when we move this to core.

### DIFF
--- a/resources/scheduled_task.rb
+++ b/resources/scheduled_task.rb
@@ -68,10 +68,10 @@ action :add do
     user               new_resource.user
     password           new_resource.password
     frequency          new_resource.frequency.to_sym
-    frequency_modifier new_resource.frequency_modifier unless %w(once on_logon onstart on_idle).include?(new_resource.frequency)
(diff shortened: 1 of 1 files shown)


resources/scheduled_task.rb
```diff
@@ -83,6 +83,25 @@
...
+  end
+
+  #
+  # not all frequencies in the windows_task resource support frequency_modifier
+  #
+  # @return [boolean]
+  #
+  def frequency_supports_frequency_modifier?
+    # these are the only ones that don't
+    !%w(once on_logon onstart on_idle).include?(new_resource.frequency)
```
phiggins: Just checking, but is `onstart` a typo? Does it need to be `on_start` to match the others?

tas50: Sadly not a typo: https://github.com/chef/chef/blob/master/lib/chef/resource/windows_task.rb#L72

PR #672, opened 2020-03-25
Initial take at a chef_client_systemd_timer resource

tas50:
Make it super simple to configure chef-client to run as a systemd-timer.

### DIFF
--- a/resources/systemd_timer.rb
+++ b/resources/systemd_timer.rb
@@ -1,0 +1,118 @@
+#
+# Cookbook:: chef-client
+# resource:: chef_client_systemd_timer
+#
+# Copyright:: 2020, Chef Software, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


resources/systemd_timer.rb
```diff
@@ -0,0 +1,123 @@
...
+
+property :job_name, String, default: 'chef-client'
+property :delay_after_boot, String, default: '1min'
+property :interval, String, default: '30min'
+
+property :splay, [Integer, String], default: 300,
+                                    coerce: proc { |x| Integer(x) },
+                                    callbacks: { 'should be a positive number' => proc { |v| v > 0 } }
+
+property :description, String, default: 'Chef Infra Client periodic execution'
```
phiggins: Should this use the `Chef::Dist` values for the product name and log directory and all that?

tas50: We're supporting chef releases before the Dist stuff was added in this cookbook.

PR #699, opened 2020-08-07
Fix systemd_timer_resource :remove action failures

tas50:
Get the underlying action right


tas50:
This is going to fail in CI since we're actually running the chef 16 resource which has the same problem. It'll resolve once we release a new client there.

Issue #627, opened 2019-05-15
Yet another Systemd Timer activation issue (edge case)


Annih:
### Cookbook version
11.2.0

### Chef-client version
12.18.31 (old one sorry, but I'm sure it is reproducible with >= 14)

### Platform Details
CentOS 7.4

### Scenario:
Reboot a server after system upgrades and have Chef-client still running every X minutes - when configured as timer.

### Steps to Reproduce:
1. Setup Chef-client as timer
2. Reboot

This is a rare edge case, I didn't manage to reproduce it on demand.  I suspect a race condition blocking the activation of the `chef-client.service` at boot time.

### Expected Result:
Chef runs every X minutes.

### Actual Result:
Chef never runs.

Timer is in weird state:
```sudo systemctl list-units chef-client.timer
UNIT              LOAD   ACTIVE SUB     DESCRIPTION
chef-client.timer loaded active elapsed chef-client periodic run

sudo systemctl list-timers chef-client.timer
NEXT LEFT LAST PASSED UNIT              ACTIVATES
n/a  n/a  n/a  n/a    chef-client.timer chef-client.service

sudo systemctl status chef-client.timer
● chef-client.timer - chef-client periodic run
   Loaded: loaded (/etc/systemd/system/chef-client.timer; enabled; vendor preset: disabled)
   Active: active (elapsed) since Tue 2019-05-14 08:06:35 UTC; 34min ago
```

Annih:
From #613
> we have had reliable activation for the most part. the one exception to that is that a few months back, possibly due to a systemd update (we're on CentOS 7), the service unit stopped activating after reboots, and the timer would not trigger until the service unit had been started. we overrode the systemd_unit[chef-client.service] resource to enable it, and thereafter once the initial chef-client.service run was triggered on boot-up, the timer would actually do its thing.

@nathwill don't you think we should enable the `chef-client.service`  even in timer mode?

Another solution could be to translate the configured interval into a `systemd.time` compatible format and use `OnCalendar`. Would it be accepted @tas50 ?

dheerajd-msys:
@tas50  We may need your input here ^.

dheerajd-msys:
@Annih We've recently made some changes in cookbook for Systemd Timer, Could you please check and confirm if it's still there?

cgunasree08:
[...]

PR #707, opened 2020-08-30
Make the onbootsec parameter to attribute

vineethp088:
This is to make the OnBootSec parameter to control via attribute when the chef-client runs as a timer service,

### Description
This is to make the OnBootSec parameter to control via attribute when the chef-client runs as a timer service.
There are cases chef-client run after the boot is elapsed, becuase of the default 1min and there should be a mechanism to control this parameter. 
Changing this parameter to hgher value seems to be fixing this and should be control it via attribute, keeping 1 min as default.

### Issues Resolved
There are cases chef-client run after the boot is elapsed, becuase of the default 1min and there should be a mechanism to control this parameter. 
Changing this parameter to higher value seems to be fixing this and should be control it via attribute, keeping 1 min as default.

### Check List

### DIFF
--- a/resources/systemd_timer.rb
+++ b/resources/systemd_timer.rb
@@ -27,7 +27,7 @@ property :description, String, default: 'Chef Infra Client periodic execution'
 
 property :user, String, default: 'root'
 
-property :delay_after_boot, String, default: '1min'
+property :delay_after_boot, String, default: '60sec'
 property :interval, String, default: '30min'
 property :splay, [Integer, String], default: 300,
                                     coerce: proc { |x| Integer(x) },
--- a/resources/systemd_timer.rb
+++ b/resources/systemd_timer.rb
@@ -27,7 +27,7 @@ property :description, String, default: 'Chef Infra Client periodic execution'
 
 property :user, String, default: 'root'
 
-property :delay_after_boot, String, default: '60sec'
+property :delay_after_boot, String, default: '1min'
 property :interval, String, default: '30min'
 property :splay, [Integer, String], default: 300,
                                     coerce: proc { |x| Integer(x) },
--- a/resources/systemd_timer.rb
+++ b/resources/systemd_timer.rb
@@ -27,7 +27,7 @@ property :description, String, default: 'Chef Infra Client periodic execution'
 
 property :user, String, default: 'root'
 
-property :delay_after_boot, String, default: '60Sec'
+property :delay_after_boot, String, default: '1min'
 property :interval, String, default: '30min'
 property :splay, [Integer, String], default: 300,
                                     coerce: proc { |x| Integer(x) },
--- a/resources/systemd_timer.rb
+++ b/resources/systemd_timer.rb
@@ -27,7 +27,7 @@ property :description, String, default: 'Chef Infra Client periodic execution'
 
 property :user, String, default: 'root'
 
-property :delay_after_boot, String, default: '1min'
+property :delay_after_boot, String, default: '60Sec'
 property :interval, String, default: '30min'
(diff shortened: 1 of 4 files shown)


resources/systemd_timer.rb
```diff
@@ -27,7 +27,7 @@
 
 property :user, String, default: 'root'
 
-property :delay_after_boot, String, default: '1min'
```
tas50, line 30: This resource is actually a backport from Chef Infra Client 16 so we need to keep the default value the same as it is in client at 1min.

dheerajd-msys:
Issue https://github.com/chef-cookbooks/chef-client/issues/627 probably be fixed by this PR

resources/systemd_timer.rb
vineethp088: We wanted to make the OnBootsec parameter as attribute, so that it can be controlled. We are seeing a problem with the chef client timer its getting elapased ( when runinng the chef client as a timer). We do not see that problem all the time/servers but we could reproduce the elapsed issue after reducing/increasing the onbootsec parameter. So if it can be controlled by node attribute it will help us fixing this issue. We can keep the default as ( 1 min/60sec) , May be I didnt get the point here about the chef16 dependecny.

phiggins: This cookbook has a systemd resource copied from Chef Infra 16, so that users of the cookbook can use older versions of Chef Infra. In order to ease transitioning to use the resource in Chef Infra 16, this needs to be defined the same as the one there: https://github.com/chef/chef/blob/master/lib/chef/resource/chef_client_systemd_timer.rb#L65-L67

The other changes in this PR are fine, but changing the default from `'1min'` to `'60sec'` needs to be reverted.

vineethp088: Sure, Reverted  let me know if this is fine. We just need that attribute to be controlled.

phiggins:
I think this is change is a good idea, but I don't think this is the right way to do it and this is complicated by what looks like a bug from confusing behavior.

* The properties `delay_after_boot`, `interval`, and `splay` are eventually used as [systemd time spans](https://www.freedesktop.org/software/systemd/man/systemd.time.html).
* `delay_after_boot` and `interval` accept arbitrary strings, so they can be a number of seconds eg `'60'` or it can include time units eg `5h30m`.
* `splay` is coerced to an integer so it is assumed to be a number of seconds.

Appending `sec` onto the value specified for `delay_after_boot` in `recipes/systemd_service.rb` could end up with incorrect values like `'5h30msec`. That also means the current behavior of interpolating `interval` is incorrect. Since bare values are interpreted as seconds, the code in the recipe should probably use the values directly:

```
      'OnBootSec' => node['chef_client']['delay_after_boot'],
      'OnUnitInactiveSec' => node['chef_client']['interval'],
      'RandomizedDelaySec' => node['chef_client']['splay'],
```

resources/systemd_timer.rb
vineethp088: Please let me know if this can be merged or need any additional modification ?

dheerajd-msys: @vineethp088 Please check https://github.com/chef-cookbooks/chef-client/pull/707#issuecomment-694400666

dheerajd-msys:
@vineethp088 Could you please make changes per suggestion [here](https://github.com/chef-cookbooks/chef-client/pull/707#issuecomment-694400666) ? then I guess we would be good to merge the PR.

Thanks

cgunasree08:
This issue or pull request is being automatically closed because the repository is being archived as part of the [Repository Standardization Initiative](https://github.com/chef-boneyard/oss-repo-standardization-2025).

PR #715, opened 2020-10-04
Simplify and correct chef-client path determination in cron recipe

### LINKED ISSUE #714: no chef-client command in cron entry
chef-client version: 13.8.5
chef-client cookbook version: 12.3.1

The cookbook is generating the following cron entry, which is misisng the chef-client in the command line.

This seems tro be a recent development.
```
# Generated by Chef. Changes will be overwritten.
LC_ALL=en_US.UTF-8

15,45 */1 * * * root /bin/sleep 88; /bin:/sbin:/usr/bin:/usr/sbin:/usr/local/bin:/usr/local/sbin --no-fork --no-color --once -c /etc/chef/client.rb > /var/log/chef/var/log/chef-client.last 2>&1
```
Here are the environment attributes
```
    "chef_client": {
      "config": {
        "verify_api_cert": false
      },
      "cron": {
        "environment_variables": {
          "LC_ALL": "en_US.UTF-8"
        },
        "hour": "*/1",
        "log_file": "/var/log/chef-client.last",
        "minute": "15,45"
      },
      "handler": {}
    },
```

tas50:
Stop using the find_chef_client helper to search for the chef-client binary and just use the node attribute that defaults to /opt/chef/bin/chef-client
Pass the correct binary attribute in the resource. This was causing the path to be set instead.

This should fix https://github.com/chef-cookbooks/chef-client/issues/714

### DIFF
--- a/recipes/cron.rb
+++ b/recipes/cron.rb
@@ -26,8 +26,6 @@ class ::Chef::Recipe
 end
 
 # libraries/helpers.rb method to DRY directory creation resources
-client_bin = find_chef_client
-node.default['chef_client']['bin'] = client_bin
 create_chef_directories
 
 # Stop any running chef-client services
@@ -56,7 +54,7 @@ when 'freebsd'
     source 'freebsd/chef-client.erb'
     owner 'root'
     group 'wheel'
-    variables client_bin: client_bin
+    variables client_bin: node['chef_client']['bin']
     mode '0755'
   end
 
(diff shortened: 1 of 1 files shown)


iambaljitdhanoa:
Hi 
@tas50 @Nirbhay1997

I have used the below scenario for node
vi /etc/crontab
then run below command
* * * * * root chef-client
:wq
The node is bootstrapped completely and when I run chef-client manually, it fetches the data from server. But the crontab file is not working. it should work automatically. Please suggest how can solve this problem

PR #716, opened 2020-10-07
Add support for service scheduling to systemd timer

axelrtgs:
Ref: https://www.freedesktop.org/software/systemd/man/systemd.exec.html#Scheduling

### DIFF
--- a/test/cookbooks/test/recipes/systemd_timer_resource.rb
+++ b/test/cookbooks/test/recipes/systemd_timer_resource.rb
@@ -2,6 +2,11 @@ chef_client_systemd_timer 'schedule chef-client to run as cron job' do
   accept_chef_license true
   daemon_options ['--run-lock-timeout 0']
   environment 'FOO' => '1', 'BAR' => '2'
+  nice 19
+  cpu_scheduling_policy 'idle'
+  cpu_scheduling_priority 1
+  io_scheduling_class 'idle'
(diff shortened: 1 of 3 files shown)


test/cookbooks/test/recipes/systemd_timer_resource.rb
```diff
@@ -2,6 +2,11 @@
   accept_chef_license true
   daemon_options ['--run-lock-timeout 0']
   environment 'FOO' => '1', 'BAR' => '2'
+  nice 19
```
ramereth, line 5: This is failing with the following:
```
undefined method `nice' for Chef::Resource::ChefClientSystemdTimer
```

axelrtgs: yeah because its testing on chef 16 which includes the resource and it takes precendence over the one in the cookbook. `chef_version_for_provides '< 16.0' if respond_to?(:chef_version_for_provides)` is in the resource which means the code in the cookbook is never truly tested and we are only testing the chef resource works. we may want to test on chef 15 so the cookbook resources also work.

FYI i had a chat with @tas50 and it looks like this doesn't do what you would expect because of forking/threading so i think i found a solution and will see to get it implemented if its ok with tim.

Not sure what to do about the tests running against chef 16 and not actually testing the cookbook code. open to ideas.

PR #717, opened 2020-10-29
chef-client stop init.d service stopping other process fixes

antima-gupta:
Signed-off-by: antima-gupta <agupta@msystechnologies.com>

### Description
- Fixed stop init.d service stopping other process issue.

### Issues Resolved
[#221](https://github.com/chef/customer-bugs/issues/221)

### Check List

### DIFF
--- a/templates/default/redhat/init.d/chef-client.erb
+++ b/templates/default/redhat/init.d/chef-client.erb
@@ -43,6 +43,7 @@ start() {
 
 stop() {
     echo -n $"Stopping $prog: "
+    ps -ef |grep $exec |grep -v grep |awk '{print $2}' > $pidfile
     killproc -p $pidfile $exec
     retval=$?
     echo


templates/default/redhat/init.d/chef-client.erb
```diff
@@ -43,6 +43,8 @@ start() {
 
 stop() {
     echo -n $"Stopping $prog: "
+    # It will pick the process id directly from the running process and adding into $pidfile.
+    ps -ef |grep $exec |grep -v grep |awk '{print $2}' > $pidfile
```
tas50, line 47: Has this been tested to confirm this works. -p is for a pidfile. Does it work if you stick a PID in there instead of a path?

antima-gupta: @tas50 Yes it has been testes and working fine. It won't work if we stick PID in there because `killproc` does not having that type of functionality. In `killproc` function there are mentioned that if you are mentioning -p then you have to give pidfile not pid.

PR #719, opened 2020-11-02
Correct the log_file_name default value in README.md

signe:
Incidental documentation change to correct the log_file_name descriptions in README.md to match what's actually used in the resources


phiggins (approved):
Nice catch!

tas50:
Thanks for the fix

PR #721, opened 2021-01-08
Expose StandardOutput and StandardError as attributes for systemd recipe

mach-kernel:
Hi! We use rsyslog for logging and have found that setting these directives to `null` will prevent duplication in journald. Unsure how to approach testing for this but consequently also think the changes are rather innocuous. Thank you for your time!

### Description

[According to this documentation](https://www.freedesktop.org/software/systemd/man/systemd-journald.service.html):

> The systemd service manager invokes all service processes with standard output and standard error connected to the journal by default. This behaviour may be altered via the StandardOutput=/StandardError= unit file settings, see systemd.exec(5) for details. The journal converts the log byte stream received this way into individual log records, splitting the stream at newline ("\n", ASCII 10) and NUL bytes.

systemd.exec pages for:
[StandardOutput](https://www.freedesktop.org/software/systemd/man/systemd.exec.html#StandardOutput=)
[StandardError](https://www.freedesktop.org/software/systemd/man/systemd.exec.html#StandardError=)

### Issues Resolved
n/a

### Check List


mach-kernel:
:wave:

dspinner (approved):

PR #739, opened 2021-07-19
[chef-client] Enable chef-client scheduled task to behave like cron, …

gholtiii:
Make chef-client scheduled task behave like cron, with a predictable start time and splay that stays consistent for a given node

### Description

The chef-client windows scheduled task uses a random delay that can be up to 25 minutes and a start time based on when the code actually runs.  Instead this change is meant to make the start time predictable, similar to a cron entry like */30, and the splay is used to compute a distinct per node value identical to how it's being done on cron (splay_sleep_time method).  This enables the next actual start time for a chef run to be predictable for a given node.

### Check List

### DIFF
--- a/resources/scheduled_task.rb
+++ b/resources/scheduled_task.rb
@@ -20,6 +20,7 @@
 chef_version_for_provides '< 16.0' if respond_to?(:chef_version_for_provides)
 
 provides :chef_client_scheduled_task
+unified_mode true
 resource_name :chef_client_scheduled_task
 
 property :task_name, String,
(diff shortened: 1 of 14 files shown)


resources/scheduled_task.rb
```diff
@@ -17,9 +17,10 @@
 # limitations under the License.
 #
 
-chef_version_for_provides '< 16.0' if respond_to?(:chef_version_for_provides)
+chef_version_for_provides '< 18.0' if respond_to?(:chef_version_for_provides)
```
jasonwbarnett: @gholtiii this was added so that the resource shipped in the core of chef infra client would be used over this resource.

https://github.com/chef/chef/blob/master/lib/chef/resource/chef_client_scheduled_task.rb

In other words, these changes should be made in https://github.com/chef/chef

gholtiii: Submitted to the core https://github.com/chef/chef/pull/11894

I'll fix this version constraint and resubmit, though it seems this cookbook should be deprecated given the above.

cgunasree08:
This issue or pull request is being automatically closed because the repository is being archived as part of the [Repository Standardization Initiative](https://github.com/chef-boneyard/oss-repo-standardization-2025).

